PR #7, opened 2021-11-05
build: archive using python archive utility

bruuuuuuuce:


.github/workflows/cd.yml
```diff
@@ -37,6 +37,11 @@ jobs:
           components: rustfmt
           override: true
           
+      - name: Setup Python
+        uses: actions/setup-python@v2
```
eaddingtonwhite, line 41: lol this action name with the prop

eaddingtonwhite (approved):

PR #16, opened 2021-11-11
feat(account): add momento signup flow to cli

bruuuuuuuce:
This commit adds a new subcommand, `account`, which is where we will put all account related cli features. The first feature that this pr implements is account signup.
```
momento account signup --email test@email.com
```

depends on https://github.com/momentohq/core-infrastructure/issues/288

### DIFF
--- a/src/commands/signup/signup.rs
+++ b/src/commands/signup/signup.rs
@@ -1,0 +1,35 @@
+use log::info;
+use reqwest::Client;
+use serde::Deserialize;
+use serde_json::json;
+
+#[derive(Deserialize, Debug)]
+struct SignupResponse {
+    message: String,
+}
+
+impl Default for SignupResponse {
+    fn default() -> Self {
+        Self {
+            message: "Unable to create token".to_string(),
+        }
+    }
+}
+
+pub async fn signup_user(email: String) {
+    let url = "https://identity.cell-external-beta-1.prod.a.momentohq.com/token/create";
+
+    let body = json!({ "email": email });
+    info!("Signing up for Momento...");
+    match Client::new().post(url).json(&body).send().await {
+        Ok(resp) => {
+            if resp.status().is_success() {
+                info!("Successfully created your token! Your token will be emailed to you shortly.")
+            } else {
+                let response_json: SignupResponse = resp.json().await.unwrap_or_default();
+                panic!("Failed to create Momento token: {}", response_json.message)
+            }
+        }
+        Err(e) => panic!("{}", e),
+    };
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -28,7 +28,7 @@ enum Subcommand {
     },
     #[structopt(about = "Configure Momento Credentials")]
     Configure {},
-    #[structopt(about = "Account Operations")]
+    #[structopt(about = "Manage Accounts")]
     Account {
         #[structopt(subcommand)]
         operation: AccountCommand,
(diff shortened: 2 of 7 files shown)


src/main.rs
```diff
@@ -114,5 +128,10 @@ async fn main() {
             }
         },
         Subcommand::Configure {} => commands::configure::configure::configure_momento().await,
+        Subcommand::Account { operation } => match operation {
+            AccountCommand::Signup { email } => {
+                commands::signup::signup::signup_user(email).await;
```
kvcache, line 133: Nit: Maybe the extra folder here doesn't really help anything. That's writing signup quite a few times in one line, has a smell 😃

src/main.rs
```diff
@@ -28,6 +28,20 @@ enum Subcommand {
...
+    #[structopt(about = "Account Operations")]
+    Account {
+        #[structopt(subcommand)]
+        operation: AccountCommand,
+    },
+}
+
+#[derive(StructOpt)]
+enum AccountCommand {
+    #[structopt(about = "Signs a user up for Momento and emails your token to the provided email")]
```
kvcache, line 40: This docstring is user facing right?

It's mixing the passive voice with direct speech / active voice. I think these messages should consistently either say what they do or tell the user what they do - in language that makes sense for an end user. Don't write the message to tell us what the command does!

I think this could be as simple as "Sign up for Momento." or "Signs you up for Momento." Idk what the preferred phrase style is or if we chose one but I think our sentences should at lease individually refrain from mixing their voice? I am also moderately opposed to using the passive voice for end user documentation of actions.

src/commands/signup/signup.rs
```diff
@@ -0,0 +1,35 @@
...
+        Self {
+            message: "Unable to create token".to_string(),
+        }
+    }
+}
+
+pub async fn signup_user(email: String) {
+    let url = "https://identity.cell-external-beta-1.prod.a.momentohq.com/token/create";
+
+    let body = json!({ "email": email });
```
kvcache, line 22: Nit: prefer serde and real structs to loose strings

src/commands/signup/signup.rs
```diff
@@ -0,0 +1,35 @@
...
+    info!("Signing up for Momento...");
+    match Client::new().post(url).json(&body).send().await {
+        Ok(resp) => {
+            if resp.status().is_success() {
+                info!("Successfully created your token! Your token will be emailed to you shortly.")
```
kvcache, line 27: How about 
Success! Your access token will be emailed to you shortly.

src/main.rs
```diff
@@ -28,6 +28,20 @@ enum Subcommand {
     },
     #[structopt(about = "Configure Momento Credentials")]
     Configure {},
+    #[structopt(about = "Account Operations")]
```
kvcache, line 31: Nit: this phrase is inconsistent with the one above it. I think it would match as "Manage accounts" or something? The one above says what it does. This one says what it is. I think we should be consistent.

kvcache (commented):
Just nits, this looks cool

src/main.rs
bruuuuuuuce: I like that rule, no passive voice

src/main.rs
bruuuuuuuce: 👃

kvcache (approved):
🧑‍🍳

bruuuuuuuce:
waiting to merge this until we deploy cacheadmin-v2 v0.23.0 (quite the mouthful) to beta environment

PR #18, opened 2021-11-15
feat: add support for multiple profiles

### LINKED ISSUE #3: Make credentials file have 400 readonly permissions

### LINKED ISSUE #14: Create config file for Momento Profiles
profiles should live in a file `~/.momento/config` and be of the format
```
[profile.default]
cache = "my-cache"
ttl = 500
```

bruuuuuuuce:
closes #3 
closes #14 

This pr also adds support to have multiple momento profiles on your computer. 

`momento configure` will now prompt you for a default cache name, as well as a default ttl. This will allow you to make get and set calls without having to specify these every time

```
momento cache get -k myKey
momento cache set -k myKey -v myValue
```

you can still optionally override either one of these values to use a different ttl or cache

```
momento cache get -k myKey --name different-cache
momento cache set -k myKey -v myValue -ttl 12345 --name another-cache
```

You can use alternative profiles by setting the environment variable `MOMENTO_PROFILE`

```
MOMENTO_PROFILE=production momento cache get -k myKey
```


src/utils/file.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use home::home_dir;
+use log::debug;
+use serde::de;
+use tokio::fs::{self, File};
+use toml::de::Error;
+
+pub fn get_credentials_file_path() -> String {
+    let momento_home = get_momento_dir();
+    return format!("{}/credentials.toml", momento_home);
```
gautamomento, line 12: Probably, we should just drop the .toml extension unless there is something in this code that requires it.

gautamomento (approved):
LGTM

Just one small comment.

src/utils/file.rs
bruuuuuuuce: discussed in slack, we are going to keep this extension for now

gautamomento (approved):

PR #25, opened 2021-11-29
fix: momento configure when user is first configuring momento

### LINKED ISSUE #23: Fix Typo
https://github.com/momentohq/momento-cli/blob/5578abe444fc8fac7f1781efbe0414c25bfd5c48/src/utils/file.rs#L29

### LINKED ISSUE #24: Remove unnecessary function
https://github.com/momentohq/momento-cli/blob/5578abe444fc8fac7f1781efbe0414c25bfd5c48/src/utils/file.rs#L78-L85

The function referenced above is unnecessary - This assertion is based on reading the documentation for `create_dir_all` which indicates: "Recursively create a directory and all of its parent components if they are missing."

https://doc.rust-lang.org/std/fs/fn.create_dir_all.html

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=36ce82190ca04d07321b69a2152ebd3d

the line below can simply be replaced with `fs::create_dir_all(path)` 
https://github.com/momentohq/momento-cli/blob/5578abe444fc8fac7f1781efbe0414c25bfd5c48/src/commands/configure/configure.rs#L26

I also do not understand why this is an async function. Cool tool - but a) why make this method by async to begin with if b) the only time you use it is when you do a .await anyway for c) a single threaded app? Eager to learn more as I learned something last time I asked this question in TS land :).

bruuuuuuuce:
closes #24 
closes #23

### DIFF
--- a/src/utils/file.rs
+++ b/src/utils/file.rs
@@ -5,7 +5,6 @@ use home::home_dir;
 use log::debug;
 use serde::de;
 use tokio::fs::{self, File};
-use toml::de::Error;
 
 pub fn get_credentials_file_path() -> String {
     let momento_home = get_momento_dir();
@@ -22,17 +21,14 @@ pub fn get_momento_dir() -> String {
     return format!("{}/.momento", home.clone().display());
 }
 
-pub async fn read_toml_file<T: de::DeserializeOwned>(path: &str) -> Result<T, Error> {
+pub async fn read_toml_file<T: de::DeserializeOwned>(path: &str) -> Result<T, String> {
     let toml_str = match fs::read_to_string(&path).await {
         Ok(s) => s,
-        Err(_) => panic!(
-            "faile to read file {}, please run 'momento configure'",
-            path
-        ),
+        Err(e) => return Err(format!("failed to read toml file: {}", e)),
     };
     match toml::from_str::<T>(&toml_str) {
         Ok(c) => Ok(c),
-        Err(e) => Err(e),
+        Err(e) => Err(format!("failed to parse toml file: {}", e)),
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/utils/file.rs
```diff
@@ -22,17 +21,14 @@ pub fn get_momento_dir() -> String {
...
 
-pub async fn read_toml_file<T: de::DeserializeOwned>(path: &str) -> Result<T, Error> {
+pub async fn read_toml_file<T: de::DeserializeOwned>(path: &str) -> Result<T, String> {
     let toml_str = match fs::read_to_string(&path).await {
         Ok(s) => s,
-        Err(_) => panic!(
-            "faile to read file {}, please run 'momento configure'",
-            path
-        ),
+        Err(e) => return Err(format!("failed to read toml file: {}", e)),
```
gautamomento, line 27: nit: Since this is user facing message, I would just drop details around format.

`Failed to read file: {}` should sufficient

src/utils/file.rs
```diff
@@ -22,17 +21,14 @@ pub fn get_momento_dir() -> String {
...
     };
     match toml::from_str::<T>(&toml_str) {
         Ok(c) => Ok(c),
-        Err(e) => Err(e),
+        Err(e) => Err(format!("failed to parse toml file: {}", e)),
```
gautamomento, line 31: ditto

src/utils/file.rs
```diff
@@ -75,12 +71,18 @@ pub async fn set_file_read_write(path: &str) -> std::io::Result<()> {
...
-    if !Path::new(path).exists() {
+pub async fn create_dir_if_not_exists(str_path: &str) {
+    let path = Path::new(str_path);
+    if !path.exists() {
         let res = fs::create_dir_all(path).await;
         match res {
-            Ok(_) => debug!("created directory {}", path),
-            Err(e) => panic!("failed to created directory {}, error: {}", path, e),
+            Ok(_) => debug!("created directory {}", str_path),
+            Err(e) => panic!("failed to created directory {}, error: {}", str_path, e),
```
gautamomento, line 80: `created` -> `create`

src/utils/file.rs
```diff
@@ -75,12 +71,18 @@ pub async fn set_file_read_write(path: &str) -> std::io::Result<()> {
...
         }
+    } else if !path.is_dir() {
+        panic!(
+            "failed to create directory, file already exists at {}",
```
gautamomento, line 84: nit: `failed to create a directory, as file with same name already exists` ?

gautamomento (commented):
Looks fine overall, though left some nit comments.

gautamomento (approved):

PR #43, opened 2022-01-31
chore: default ttl to 600 seconds if empty

### LINKED ISSUE #29: Default TTL should not be 0
Currently, the default TTL is 0, but in reality it's more like 1 second. Let's change the default ttl to something like 600s (5 min)

danielamiao:
closes #29


bruuuuuuuce (approved):
:shipit:

PR #46, opened 2022-02-02
feat: pretty print cli error messages

### LINKED ISSUE #41: Use `clap` instead of `structopt`
Use clap instead of structopt which is abandoned and has bugs

### LINKED ISSUE #42: Improve error formatting
- Better error log formatting (not panics like below)
    - `[2022-01-27T18:26:29Z ERROR momento] panicked at 'Cache not found', src/commands/cache/cache.rs:17:19`

bruuuuuuuce:
closes #42 
closes #41 
Screen Shot 2022-02-01 at 4 25 15 PM

Right now, I just implemented 1 error type called `CliError`, but I figured that we could pretty easily add more custom error types down the line, and each can be formatted appropriately

### DIFF
--- a/src/error/mod.rs
+++ b/src/error/mod.rs
@@ -1,0 +1,19 @@
+use std::fmt;
+
+use colored::Colorize;
+
+pub struct CliError {
+    pub(crate) msg: String,
+}
+
+impl fmt::Debug for CliError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}: {:#?}", "ERROR".red().bold(), self.msg.red())
+    }
+}
+
+impl fmt::Display for CliError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}: {}", "ERROR".red().bold(), self.msg.red())
+    }
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-use std::panic;
+use std::{panic, process::exit};
 
 use clap::StructOpt;
 use env_logger::Env;
@@ -162,6 +162,8 @@ async fn main() {
     }));
 
     if let Err(e) = entrypoint().await {
-        eprintln!("{}", e)
+        eprintln!("{}", e);
+        exit(1)
     }
+    exit(0)
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -165,5 +165,4 @@ async fn main() {
         eprintln!("{}", e);
         exit(1)
     }
-    exit(0)
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,15 +1,17 @@
 use std::panic;
 
+use clap::StructOpt;
 use env_logger::Env;
+use error::CliError;
 use log::error;
-use structopt::StructOpt;
 use utils::user::get_creds_and_config;
 
 pub mod commands;
 mod config;
+pub mod error;
 mod utils;
 
-#[derive(StructOpt)]
+#[derive(Debug, StructOpt)]
 #[structopt(about = "CLI for Momento APIs")]
 struct Momento {
     #[structopt(name = "verbose", global = true, long)]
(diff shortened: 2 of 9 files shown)


src/main.rs
```diff
@@ -113,46 +110,58 @@ async fn main() {
...
+#[tokio::main]
+async fn main() {
+    panic::set_hook(Box::new(move |info| {
+        error!("{}", info);
+    }));
+
+    if let Err(e) = entrypoint().await {
+        eprintln!("{}", e)
+    }
 }
```
kvcache, line 167: And then exit with code `0`? Is this accidental or intentional?

kvcache (commented):
I'm not sure `CLIError` makes sense in every case here, but replacing `panic!` with messages is quite desirable.

src/main.rs
bruuuuuuuce: I didn't think I needed an exit 0, does this just hang in the background if we don't exit 0?

kvcache: No I mean you print error then exit. If you do `echo $?` after an erroring `momento` invocation you'll see `0`. I think if `momento` ends with `Err` then `$?` should be nonzero.

bruuuuuuuce:
> I'm not sure `CLIError` makes sense in every case here, but replacing `panic!` with messages is quite desirable.

Are you thinking something along the lines of 
- ServiceError
- IoError
- BadRequestError
- PermissionError

src/main.rs
bruuuuuuuce: ah yes makes sense!

kvcache: ```momento cache get --key matt_doesnt_exist && echo this message should not print because momento should have exited with something other than 0```

Like this:
```
➜  proto git:(main) ✗ momento cache get --key matt_doesnt_exist && echo this message should not print because momento should have exited with something other than 0
[2022-02-02T01:26:12Z INFO  momento::commands::cache::cache] getting key: matt_doesnt_exist from cache: roflmao
[2022-02-02T01:26:13Z ERROR momento] panicked at 'Cache not found', src/commands/cache/cache.rs:17:19
➜  proto git:(main) ✗ 
```

Note that the shell is told that `momento` failed. That's desirable. I ¿think¿ if you do this you will see your red ERROR message and then "this message should not print[...]"

kvcache:
> > I'm not sure `CLIError` makes sense in every case here, but replacing `panic!` with messages is quite desirable.
> 
> Are you thinking something along the lines of
> 
>     * ServiceError
> 
>     * IoError
> 
>     * BadRequestError
> 
>     * PermissionError

Yes - but you'll end up rolling them all up to CLIError anyway because you won't do much with them today. If you want to, I've used https://docs.rs/thiserror/latest/thiserror/ to great success. You can even roll up AWS error types under a wrapper error type you control.

This is not blocking this pr.

src/main.rs
```diff
@@ -113,46 +110,60 @@ async fn main() {
...
+async fn main() {
+    panic::set_hook(Box::new(move |info| {
+        error!("{}", info);
+    }));
+
+    if let Err(e) = entrypoint().await {
+        eprintln!("{}", e);
+        exit(1)
+    }
+    exit(0)
```
kvcache, line 168: nit: letting the function end already naturally does exit code 0 / this is unnecessary.

kvcache (approved):
nice

src/error/mod.rs
```diff
@@ -0,0 +1,19 @@
+use std::fmt;
+
+use colored::Colorize;
+
+pub struct CliError {
```
cprice404, line 5: you should check out the `thiserror` crate and see if you like any of the stuff it gives you.

Also check out the `anyhow` crate, it gives you a way to bubble up multiple error types.

You may not actually need either of those right now but just wanted to let you know about them if you haven't seen them yet; I'm planning on switching `mm` over to those soon.

cprice404 (approved):
This is so awesome!  I've been really wanting to do this in `mm` but haven't had time yet.

panic

PR #53, opened 2022-02-25
chore: Add JP translation for README and CONTRIBUTING

poppoerika:
Added JP translation for both `README` and `CONTRIBUTING`

@danielamiao - It seems like the example has table of contents but since this README is short I'm not sure if it's needed. 
I skipped it for now.

### DIFF
--- a/CONTRIBUTING.ja.md
+++ b/CONTRIBUTING.ja.md
@@ -1,16 +1,22 @@
 ### 開始方法
 
-1. git submodule init
-1. git submodule sync
-1. git submodule update --recursive --remote
+```
+git submodule init
+git submodule sync
+git submodule update --recursive --remote
+```
 
 ## ビルド
 
-1. cargo build
+```
+cargo build
+```
 
 ## テスト
 
-1. cargo test
+```
+cargo test
+```
 
 ## デプロイ
 
--- a/CONTRIBUTING.ja.md
+++ b/CONTRIBUTING.ja.md
@@ -1,0 +1,18 @@
+### 開始方法
+
+1. git submodule init
+1. git submodule sync
+1. git submodule update --recursive --remote
+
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.ja.md
```diff
@@ -0,0 +1,18 @@
+### 開始方法
+
+1. git submodule init
```
danielamiao, line 3: This line and below all have numbers at the beginning, you should delete it so users can copy the entire code block to terminal without having to delete the numbers

README.ja.md
```diff
@@ -0,0 +1,43 @@
+## 必要条件
```
danielamiao, line 1: I like the sample README had a line at the top stating other languages are available, linking to each README version. Could we add it here too? That way, if a Japanese user lands on the English README by accident they can easily click on the link to find the Japanese README

danielamiao: Like this line 

Read this in other languages: [English](https://github.com/tiimgreen/github-cheat-sheet/blob/master/README.md), [한국어](https://github.com/tiimgreen/github-cheat-sheet/blob/master/README.ko.md), [日本語](https://github.com/tiimgreen/github-cheat-sheet/blob/master/README.ja.md), [简体中文](https://github.com/tiimgreen/github-cheat-sheet/blob/master/README.zh-cn.md), [正體中文](https://github.com/tiimgreen/github-cheat-sheet/blob/master/README.zh-tw.md).

danielamiao (commented):
Agree table of content is not needed, a few other comments!

README.ja.md
poppoerika: Added to the original ‘README.md`! 
I can add the same for the Japanese README as well.

danielamiao (approved):

PR #54, opened 2022-03-01
feat: create default cache when configuring profile

bruuuuuuuce:
Screen Shot 2022-03-01 at 12 07 11 PM

Do we want to show them that we are creating a cache from them? Or should we hide the logging

### DIFF
--- a/src/commands/cache/cache.rs
+++ b/src/commands/cache/cache.rs
@@ -11,20 +11,20 @@ async fn get_momento_instance(auth_token: String) -> Result<SimpleCacheClient, C
 }
 
 pub async fn create_cache(cache_name: String, auth_token: String) -> Result<(), CliError> {
-    info!("create cache called");
+    info!("creating cache...");
     let mut momento = get_momento_instance(auth_token).await?;
     match momento.create_cache(&cache_name).await {
-        Ok(_) => info!("created cache {}", cache_name),
+        Ok(_) => (),
         Err(e) => return Err(CliError { msg: e.to_string() }),
     };
     Ok(())
 }
 
 pub async fn delete_cache(cache_name: String, auth_token: String) -> Result<(), CliError> {
-    info!("delete cache called");
+    info!("deleting cache...");
     let mut momento = get_momento_instance(auth_token).await?;
     match momento.delete_cache(&cache_name).await {
-        Ok(_) => info!("deleted cache {}", cache_name),
+        Ok(_) => (),
         Err(e) => return Err(CliError { msg: e.to_string() }),
     };
     Ok(())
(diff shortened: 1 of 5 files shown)


bruuuuuuuce:
for some reason I am not able to transfer this issue over https://github.com/momentohq/dev-eco-issue-tracker/issues/38 to this repo

mjindalmotiv8:
I think instead of `create cache called` - maybe we can change it to `creating cache`

mjindalmotiv8:
What is the experience when create cache fails?

danielamiao:
@bruuuuuuuce probably cuz it's a private => public repo. What happens if the cache name already exists?

bruuuuuuuce:
> @bruuuuuuuce probably cuz it's a private => public repo. What happens if the cache name already exists?

If the api call fails you will just see this
Screen Shot 2022-03-01 at 12 15 46 PM

danielamiao:
@bruuuuuuuce looks like the same screenshot?

bruuuuuuuce:
> @bruuuuuuuce looks like the same screenshot?

So if the cache creates successfully you will see

```
create cache called
created cache <name>
```

if it fails you will just see
```
create cache called
```

bruuuuuuuce:
I honestly think it might be better to just not even show anything, but this was the simpler change so i put this up first.

danielamiao:
| I honestly think it might be better to just not even show anything, but this was the simpler change so i put this up first.

I think it's definitely better to show nothing if the call fails - user will run into the error on the next get/set command, which is what happens now. Currently, the message is confusing to show - it makes me think it created it for me successfully

eaddingtonwhite:
I would just show nothing at all then if were not going to show failures. we should not show that we are doing this or if it succeeds.

Maybe in CLI then on get and set we can handle Cache not found error and prompt them with command to create cache or even try to do it lazily for them if we detect first call fails.

danielamiao:
> Maybe in CLI then on get and set we can handle Cache not found error and prompt them with command to create cache 

I like this, but we can do this in a separate PR as an improvement to handle cases where creates fails during `momento configure`

bruuuuuuuce:
Screen Shot 2022-03-08 at 9 38 55 AM

danielamiao:
LGTM, added @poppoerika to review as well since this might conflict with her other PR after getting merged.

poppoerika (commented):
Looks good to me as well. 
Updates to my pr won't be too much!

danielamiao:
@bruuuuuuuce one more request:
- When calling create cache, change message to `creating cache...`, as @mjindalmotiv8 suggested above.
- After the call returns, either print `default cache successfully created` or `default cache already exists` depending on if the response was OK or not

Issue #58, opened 2022-03-07
Update existing config file without clobbering any data 


poppoerika:
Port over "update config without clobbering" functionality from python to Rust, then add support to `momento configure` to enable editing config via CLI.

Master ticket: https://github.com/momentohq/momento-cli/issues/40

PR #59, opened 2022-03-08
feat: Add support for multiple profiles and only allow new credentials/config files creations

### LINKED ISSUE #56: Add support for momento config file multiple profiles
Add support for momento config file to support multiple profiles (this unblocks an immediate demand to store multiple tokens for multiple regions, including myself). For instance, I want to be able to do`momento --profile public-us-east-1 cache list`

Master ticket: https://github.com/momentohq/momento-cli/issues/40

### LINKED ISSUE #57: Throw an error if a user tries to run `momento configure` with existing config file
For now, `momento configure` supports onboarding new users only, i.e. it only works if you do not have an existing momento config file. This command basically creates the config file and stores a token, cache name, TTL for default named profile.

If a user with an existing momento config file tries running `momento configure` , it will throw an error and print something like "Existing config file detected, please edit $HOME/.momento/config directly to add or modify profiles".

Master ticket: https://github.com/momentohq/momento-cli/issues/40

poppoerika:
Conducted local test scenarios:
- Run `momento cache list --profile <profile>` to check if we can specify a profile. When `--profile` is not specified, `default` profile will be used.
image
<br>

- Run `momento configure` without credentials or config files -> create new credentials and config files with `[default]` section in INI format.
<br>

- Run `momento configure` only with existing credentials file -> create a new config file with [default]` section in INI format and throw an error saying that the credentials file already exists.
image
<br>

- Run `momento configure` only with existing config file -> create a new credentials file with [default]` section in INI format and throw an error saying that the config file already exists.
image
<br>

- Run `momento configure` with both existing credentials and config files, throw an error saying that both credentials and config files exist.
image

Closes #56 
Closes #57

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -197,6 +197,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "configparser"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "06821ea598337a8412cf47c5b71c3bc694a7f0aed188ac28b836fab164a2c202"
+
+[[package]]
 name = "core-foundation"
 version = "0.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -667,6 +673,7 @@ dependencies = [
  "assert_cmd",
  "clap",
  "colored",
+ "configparser",
  "env_logger",
  "home",
  "log",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ serde = {version = "1.0", features = ["derive"] }
 maplit = "1.0.2"
 reqwest = { version = "0.11", features = ["json", "rustls-tls"], default-features = false }
 momento = { path = "./client-sdk-rust" }
+configparser = "3.0.0"
 
 [dev-dependencies]
 assert_cmd = "2.0.2"
--- a/src/commands/configure/configure.rs
+++ b/src/commands/configure/configure.rs
@@ -1,19 +1,18 @@
 use log::debug;
-use serde::{de::DeserializeOwned, Serialize};
 
+use configparser::ini::Ini;
 use tokio::{
     fs,
     io::{self, AsyncBufReadExt, AsyncWriteExt, BufReader},
 };
 
 use crate::{
-    config::{Config, Credentials, Profiles},
+    config::{Config, Credentials},
     error::CliError,
     utils::{
         file::{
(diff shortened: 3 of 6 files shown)


danielamiao:
@bruuuuuuuce Can we combine `config` and `credentials` file into one? Just `config` and throw the token into the profile in `credentials`? I remember you had a reason to keep them separate initially but I think maybe we've changed our mind since then? It's easier to keep track of just 1 config file.

poppoerika:
@bruuuuuuuce / @danielamiao - Is it okay to turn off the integration test for now since it'll always fail with this updates?

bruuuuuuuce:
> @bruuuuuuuce Can we combine `config` and `credentials` file into one? Just `config` and throw the token into the profile in `credentials`? I remember you had a reason to keep them separate initially but I think maybe we've changed our mind since then? It's easier to keep track of just 1 config file.

The reasoning to separate these into 2 files was for security. We wanted to treat the credentials file as sort of a pem key with 400 permissions vs the config file would have more open permissions cc @kvcache.


 > @bruuuuuuuce / @danielamiao - Is it okay to turn off the integration test for now since it'll always fail with this updates?
 
Yea I am ok with this for now

kvcache:
> > @bruuuuuuuce Can we combine `config` and `credentials` file into one? Just `config` and throw the token into the profile in `credentials`? I remember you had a reason to keep them separate initially but I think maybe we've changed our mind since then? It's easier to keep track of just 1 config file.
> 
> The reasoning to separate these into 2 files was for security. We wanted to treat the credentials file as sort of a pem key with 400 permissions vs the config file would have more open permissions cc @kvcache.

Yeah, consider `~/.ssh/` and `~/.aws/` as good prior art on this. Credentials and config profiles aren't the same things.

> > @bruuuuuuuce / @danielamiao - Is it okay to turn off the integration test for now since it'll always fail with this updates?
> 
> Yea I am ok with this for now

🤔 let me check my understanding: Are we planning to ship a `momento` update that is untested?

poppoerika:
@kvcache - I can certainly add different integration tests to verify behaviors of this pr in a follow-up pr. 
I wanted to merge this pr first before it gets too big if it's okay.
Also, @danielamiao and I were wondering if the approach I'm implementing in this pr seems reasonable until I start working on https://github.com/momentohq/momento-cli/issues/58.
We believe that this approach is quick and let users specify a profile to improve their CLI experience.

kvcache (approved):
yeah seems reasonable

PR #60, opened 2022-03-09
feat: impose clippy linter rules

kvcache:
Adds `cargo clippy --all-targets --all-features -- -D warnings` to ci.yml and fixes every violation of every current Clippy lint rule in `momento`. This is maximum-strength Clippy; we're opting into the most pedantic ruleset they offer (as far as I know, but without experimental rules - those aren't intended to be included).

Clippy updates, so we may run into future changes that break on PR for unrelated reasons. I think that is a small price to pay for a linter that keeps our code idiomatic and consistent but feel free to opine on that.

- fix: remove useless format!() invocations
- remove unnecessary returns
- fix needless borrows
- remove useless clone that duplicates memory without using the data
- fix useless match - add logging and DRY code
- fix module inception (do not name modules the same as the module in which they reside)
- remove unused Unit return types
- add clippy to pull_request github workflow


bruuuuuuuce (approved):
:shipit:

cprice404 (approved):

PR #64, opened 2022-03-09
chore: update README

danielamiao:
Added some instructions to view and specify region upon sign up - tried to keep it short / frictionless but open to feedback!

@poppoerika would appreciate if you can push up the Japanese translated version to reflect these changes.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -56,7 +56,3 @@ momento cache get --key key --name example-cache
 ## Momento CLI リポジトリへの貢献について
 
 もし Momento CLI リポジトリへの貢献に興味がありましたら、こちらの[貢献ガイド](./CONTRIBUTING.ja.md)のご参考をお願いします。
-
-```
-
-```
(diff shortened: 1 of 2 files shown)


rlinehan (approved):
I think this looks good to me!

We were talking about also adding a note about old versions getting errors on `account signup` and users should upgrade. Should we do that now or hold off for another PR?

danielamiao:
@rlinehan yes good point we should probably just add it now I'll push up an edit

poppoerika:
@danielamiao / @rlinehan - I'd like to add Japanese translation for this updates to this branch, so could you hold off merging it?

danielamiao:
@poppoerika Thanks! I just pushed up an additional change to add the message Ruth suggested.

rlinehan (dismissed):
English language changes look good to me!

poppoerika (approved):

poppoerika:
@danielamiao JP version is updated.

PR #70, opened 2022-03-15
fix: Add Github Token to Create version update pull request step

poppoerika:


.github/workflows/cd.yml
```diff
@@ -64,6 +64,7 @@ jobs:
           pr_label: "automation"
           pr_draft: false
           pr_allow_empty: true
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
danielamiao, line 67: how did builds work without this before?

poppoerika: We didn't have [this step](https://github.com/momentohq/momento-cli/blob/main/.github/workflows/cd.yml#L57) to create a pull request before my pr got merged.
It failed at this step since `secrets.GITHUB_TOKEN` was not provided..

bruuuuuuuce (approved):
:shipit:

PR #74, opened 2022-03-15
feat: Improve `momento configure`

### LINKED ISSUE #35: `Default cache` confusion in `configure`
This might actually be better as two separate issues, but I wasn't quite sure because I think the overarching theme is how exactly should "configure" work for the `default cache`? I also don't think either of these are particularly high priority, just things I noticed when trying out the CLI:

#### 1. When running `momento configure` it asks for `Default Cache`. It doesn't complain if you don't provide anything:

```
$ momento configure
Token [****]: <token>
Default Cache:
Default Ttl Seconds [0]:
```

However, if you then try to do `set` with this default, you get an error:

```
$ momento cache set --key cat --value halla
[2022-01-12T01:39:05Z INFO  momento::commands::cache::cache] setting key: cat into cache:
[2022-01-12T01:39:06Z ERROR momento] panicked at 'Permission denied', src/commands/cache/cache.rs:17:19
```

At the very least I would expect a better error message here than `Permission denied`? Like, if you don't have a default cache set then it prompts that you must use `--name`?

#### 2. I sort of expected that when I set my "default cache" that that cache would be created for me
This means that my initial workflow looks like:

1. run `momento configure` and set my `default cache`
2. remember what I set my default cache to and then create it with `momento cache create`
3. then I can run `momento cache set` and it won't error

It might be a slightly nicer onboarding experience if the tool created my `default cache` for me when I set it up. 

Alternatively, I could see something like maybe you can run pieces of `momento configure` separately - like you start with `momento configure auth` so that all you're setting is your auth token, and you can come back later to set a default cache if you want to?

### LINKED ISSUE #36: sanitize inputs from momento configure
* `momento configure`
* triple-click select auth token from signup email
* paste into `momento configure`

**expected:** ~/.momento/credentials.toml is valid.
**observed:** ~/.momento/credentials.toml stores tokens without stripping whitespace `token = " eyJhbGciOiJIU[...]`

### LINKED ISSUE #37: `configure` clobbers previous credentials
I already signed up for `momento`. The email told me to run `momento configure` and that clobbered my default profile's credential.

Profiles need to be handled differently. We need to prompt during configure:
```
Profile name [us-east-1 (with a suffix if there is already a profile named us-east-1)]: 
```

and then store the configuration in that properly named profile.

Then **_default_** is a different thing altogether, where you assign the profile you want to use by default like `default = "us-east-1"`. This is non-destructive and you should then prompt:

```
Set this profile to be the default? [Y/n]:
```

### LINKED ISSUE #38: `configure` clobbers comments
I have comments in my momento credentials file. They do not survive the `momento configure` command the signup email tells me to run.

### LINKED ISSUE #40: Fix how `momento configure` interacts with config file
The config file should:

- Supports default profile with a token, default cache name, and default TTL
- Supported named profiled with a token, optional cache name, and TTL

On command line, support passing in named profile via env var:

```
# use default profile
momento cache get --key

# use named profile
MOMENTO_PROFILE=iad-region momento cache get --key
```

- [x] Add support for momento config file to support multiple profiles (this unblocks an immediate demand to store multiple tokens for multiple regions, including myself). For instance, I want to be able to do momento --profile public-us-east-1 cache list 
https://github.com/momentohq/momento-cli/issues/56

- [x] For now, momento configure supports onboarding new users only, i.e. it only works if you do not have an existing momento config file. This command basically creates the config file and stores a token, cache name, TTL for default named profile.
If a user with an existing momento config file tries running momento configure , it will throw an error and print something like "Existing config file detected, please edit $HOME/.momento/config directly to add or modify profiles".
https://github.com/momentohq/momento-cli/issues/57

- [ ] We will port over "update config without clobbering" functionality from python to Rust, then add support to momento configure to enable editing config via CLI.
https://github.com/momentohq/momento-cli/issues/58

### LINKED ISSUE #68: Dont let user enter empty cache name on configure.
Should not just continue configure and then fail creating cache if we dectect user used empty cache name. This is really easy to hit when copying and pasting in a JWT to use if it has a new line on accident at end. Then it fails and no easy way to go back. Should fail upfront.

Screen Shot 2022-03-14 at 3 18 47 PM

poppoerika:
Locally tested the following:

- `momento configure` to create new credentials and config files
- `momento configure` to update existing default profile without clobbering data in existing credentials and config files
- `momento configure -p test-profile` to add a new profile named `test-profile` without clobbering data in existing credentials and config files
- `momento configure -p test-profile` to update `test-profile` data without clobbering data in existing credentials and config files
- Tried to hit enter to skip the `Default Cache:` prompt without entering anything at initial configuration ->  cache name , `default-cache` is automatically set to the default's cache
- Tried to enter `  token` for `Token:`  and `   cache` for `Default Cache:` with whitespaces in front of the values -> both credentials and config files remove the whitespaces in front of those values

Main ticket:
Closes #40

Sub tickets:
Closes #37
Closes #38
Closes #35
Closes #36  
Closes #68


danielamiao:
> Tried to hit enter to skip the Default Cache: prompt without entering anything -> ERROR: No default cache is provided. is printed and exited

I think Ellery requested the same in https://github.com/momentohq/momento-cli/issues/68, I left a comment in there but I think it's actually better experience to just have a default cache name [default-cache] similar to the TTL prompt, instead of err-ing user out on empty. This removes onboarding friction if the user is just trying to get started asap.

danielamiao:
I also just took a cursory glance through, there is definitely a lot of new code, great work! I think you can probably pull some of the logic out into helper functions and provide unit tests for them. This is especially true for parsing an existing profile and updating it without clobbering comments or existing profiles.

poppoerika:
> > Tried to hit enter to skip the Default Cache: prompt without entering anything -> ERROR: No default cache is provided. is printed and exited
> 
> I think Ellery requested the same in https://github.com/momentohq/momento-cli/issues/68, I left a comment in there but I think it's actually better experience to just have a default cache name [default-cache] similar to the TTL prompt, instead of err-ing user out on empty. This removes onboarding friction if the user is just trying to get started asap.

Tested locally:
Tried to hit enter to skip the `Default Cache:` prompt without entering anything at initial configuration -> cache name , `default-cache` is automatically set to the default's cache

eaddingtonwhite:
@poppoerika been testing locally looking great. One super nit but is it possible to put new lines inbetween profiles in `config` file so its little easier to read. Same with `credentials` file.

Right now looks like 
```ini
cat ~/.momento/config
[default]
ttl=600
cache=default
[test-profile]
cache=default-cache
ttl=600
[test-profile-2]
cache=test-cache
ttl=600
```

want

```ini
cat ~/.momento/config
[default]
ttl=600
cache=default

[test-profile]
cache=default-cache
ttl=600

[test-profile-2]
cache=test-cache
ttl=600
```

src/commands/configure/configure_cli.rs
```diff
@@ -127,30 +119,311 @@ async fn add_profile_to_credentials(
...
+                                    ..profile_line_num_array[counter + 1]
+                                {
+                                    // Check if line is not a comment or profile
+                                    if !line_array[n].starts_with('#')
+                                        && !line_array[n].starts_with('[')
+                                    {
+                                        let line_len = line_array[n].len();
+                                        // Replace value after "token="
+                                        line_array[n].replace_range(
+                                            6..line_len,
```
eaddingtonwhite, line 196: what is 6? Maybe make a constant so easy tell what this is for?

src/commands/configure/configure_cli.rs
```diff
@@ -127,30 +119,311 @@ async fn add_profile_to_credentials(
...
+                                        );
+                                    }
+                                }
+                            } else {
+                                // Case where profile_name is at the beginning or at the middle of profile_line_num_array
+                                for n in profile_line_num_array[counter]
+                                    ..profile_line_num_array[counter + 1]
+                                {
+                                    // Check if line is not a comment or profile
+                                    if !line_array[n].starts_with('#')
```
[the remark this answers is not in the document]
poppoerika: Thank you for the feedback, Ellery!
I agree that we should reduce the code duplication. As discussed on Slack, I'll create something like `ini_config.rs` under the `utils` submodule.

src/commands/configure/configure_cli.rs
poppoerika: 6 indicates the position right after `=` to start putting token or cache values. 
But it should be a constant as you mentioned or it should be more dynamic based on the length of the keys (`token`, `cache` or `ttl`)
I will update that.

poppoerika:
> One super nit but is it possible to put new lines inbetween profiles in config file so its little easier to read. Same with credentials file.

@eaddingtonwhite Hmm, not sure if we want to do that... since we don't want to clobber existing data if there is any.
For example if credentials file contains comments, we can't tell which one belongs to which profile or if they are just general comments.
```
#this is my credentials file
#it contains two profiles with tokens
[default]
token=dfdsuifhfheifuew
#this profile contains ap-northeast-1 token
[ap-northeast-1]
token=kjfiudheiuhiewfeifew
#end of confidentials file
```

If there is any new line between, then it should be kept after adding a new profile or updating existing profiles.

eaddingtonwhite:
@poppoerika can we just always enter new line above the new `[profile]` block when we insert it in for first time? If its simpler to not have to worry about it on update I think its ok to leave it how it is now so we dont mess up a comment above block the user added manually if that's easier to start.

danielamiao:
@poppoerika another NIT, right now the default cache name isn't displayed to the user like default TTL is:

```
➜  debug git:(feat/configureProfile) ./momento configure -p gcp-us-east-1
Token: <user paste token>
Default Cache:
Default Ttl Seconds [600]:
[2022-03-16T22:49:05Z INFO  momento::commands::cache::cache_cli] creating cache...
```
It should be:
```
➜  debug git:(feat/configureProfile) ./momento configure -p gcp-us-east-1
Token: <user paste token>
Default Cache [default-cache]:
Default Ttl Seconds [600]:
[2022-03-16T22:49:05Z INFO  momento::commands::cache::cache_cli] creating cache...
```

src/commands/configure/configure_cli.rs
```diff
@@ -118,39 +123,207 @@ async fn prompt_user_for_config(profile_name: &str) -> Result<Config, CliError>
...
+                } else {
+                    // If profile_name already exists, update token value
+                    let line_num_of_existing_profile =
+                        find_existing_profile_start(line_array.clone(), profile_name);
+                    match file_types.clone() {
+                        FileTypes::Credentials(cr) => {
+                            update_profile_values(
+                                profile_line_num_array,
+                                line_num_of_existing_profile,
+                                &mut line_array,
```
eaddingtonwhite, line 196: one thing see here is we pass down `line_array` (`file_contents`) down to `update_profile_values` and then w/ rust I guess we have to explicitly say it can be mutated w/ `&mut` which kinda neat makes you specify that. I generally like to keep my code following more of immutable data structures pattern as my default unless something is extremely performance sensitive need pass around shared reference to something keep reusing. I find this leads to the safest most readable code usually.

So for this function here `update_profile_values` could we instead just have it return an updated array of strings we can assign to new variable `updated_file_contents`.  So might look like
```rust
updated_file_contents = update_profile_values(
    profile_line_num_array,
    existing_profile_starting_line_num,
    file_contents,
    FileTypes::Config(cf),
);

File: src/commands/configure/configure_cli.rs [hunk 4]
```diff
@@ -118,39 +123,207 @@ async fn prompt_user_for_config(profile_name: &str) -> Result<Config, CliError>
...
+                            push_to_line_array(
+                                &mut line_array,
+                                FileTypes::Config(cf),
+                                profile_name,
+                            );
+                        }
+                    }
+                } else {
+                    // If profile_name already exists, update token value
+                    let line_num_of_existing_profile =
```
eaddingtonwhite, line 189: would call this `existing_profile_starting_line_num`

src/utils/ini_config.rs
```diff
@@ -0,0 +1,122 @@
...
+    ini_map.set(profile_name, "cache", Some(config.cache));
+    ini_map.set(profile_name, "ttl", Some(config.ttl.to_string()));
+    match ini_write_to_file(ini_map, config_file_path).await {
+        Ok(_) => Ok(()),
+        Err(e) => Err(e),
+    }
+}
+
+pub fn update_profile_values(
+    profile_line_num_array: Vec<usize>,
```
eaddingtonwhite, line 45: maybe we can call this `existing_profile_line_numbers`

src/utils/ini_config.rs
```diff
@@ -0,0 +1,122 @@
...
+    line_num_of_existing_profile: usize,
```
eaddingtonwhite, line 46: maybe call `existing_profile_starting_line_num`

src/utils/ini_config.rs
```diff
@@ -0,0 +1,122 @@
...
+    line_array: &mut [String],
```
eaddingtonwhite, line 47: call `file_contents` and make immutable `[String]`

src/commands/configure/configure_cli.rs
```diff
@@ -118,39 +123,207 @@ async fn prompt_user_for_config(profile_name: &str) -> Result<Config, CliError>
...
+                    Ok(_) => {}
+                    Err(e) => return Err(e),
+                }
+            }
+        }
+    } else {
+        // If file already exists, figure out any profiles exist in the file
+        set_file_read_write(path).await.unwrap();
+        let file = open_file(path).await.unwrap();
+        let mut line_array = read_file_contents(file).await;
```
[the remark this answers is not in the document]
poppoerika: Updated from `line_array` to `file_contents` in all the files where the variable name is used.

poppoerika:
@danielamiao @eaddingtonwhite 
I updated `README.md` as well as `README.js.md` to include how to add a new profile in addition to `default` profile, how to update any existing profiles, and how to use a specific profile when calling cache commands.
Please take a look and let me know what you think!

danielamiao (commented):
README updates look good to me!

eaddingtonwhite (approved):
This is looking good to me. Nice work on this! I'm still pretty new to rust so not sure on best practices with some of the syntax here if better way to do in rust. At high level though this looks good. Might want to have @kvcache take a quick pass as well.

README.md
```diff
@@ -25,24 +27,50 @@ momento account signup --email <TYPE_YOUR_EMAIL_HERE>
...
+momento configure --profile new-profile
+```
+
+This will prompt you the same as above and save them to be reused as a part of your `new-profile` profile.
+
+<br>
+
+### Update existing configuration
+
+To update your desired profile, use the same command as above.
```
kvcache, line 54: Nice

README.md
```diff
@@ -25,24 +27,50 @@ momento account signup --email <TYPE_YOUR_EMAIL_HERE>
...
 momento cache create --name example-cache
 momento cache set --key key --value value --ttl 100 --name example-cache
 momento cache get --key key --name example-cache
 ```
 
+You can also specify your desired profile.
+
+```
+# use new-profile
+momento cache create --name example-cache --profile new-profile
```
kvcache, line 69: Nice

src/commands/configure/configure_cli.rs
```diff
@@ -118,39 +123,207 @@ async fn prompt_user_for_config(profile_name: &str) -> Result<Config, CliError>
...
-    ini_map.set(profile_name, "ttl", Some(config.ttl.to_string()));
-    write_config_file(ini_map, config_file_path)
+fn find_profile_start(file_contents: Vec<String>) -> Option<Vec<usize>> {
+    let mut counter = 0;
+    let mut profile_counter;
+    let line_array_len = file_contents.len();
+    let mut profile_start_line_num_array: Vec<usize> = Vec::new();
+    while counter < line_array_len {
+        let line = file_contents[counter].trim();
+        if line.starts_with('[') && line.ends_with(']') {
```
kvcache, line 233: Some of this is pretty fragile, but it beats what we have! Probably would have preferred regular expressions but it's a-okay.

src/utils/ini_config.rs
```diff
@@ -0,0 +1,179 @@
...
+    existing_profile_line_numbers: Vec<usize>,
+    existing_profile_starting_line_num: usize,
+    file_contents: Vec<String>,
+    file_types: FileTypes,
+) -> Vec<String> {
+    let num_of_profiles = existing_profile_line_numbers.len();
+    let file_contents_len = file_contents.len();
+    let mut updated_file_contents: Vec<String> = Vec::new();
+    for (counter, line_num) in existing_profile_line_numbers.iter().enumerate() {
+        #[allow(clippy:needless_range_loop)]
```
kvcache, line 54: Why disable this lint instead of addressing it?

src/utils/ini_config.rs
```diff
@@ -0,0 +1,179 @@
...
+        if existing_profile_starting_line_num == *line_num {
+            // Case where profile_name is the only or last item in existing_profile_line_numbers
+            if counter == num_of_profiles - 1 {
+                #[allow(clippy::needless_range_loop)]
```
kvcache, line 58: Here too

src/utils/ini_config.rs
```diff
@@ -0,0 +1,179 @@
...
+                                    n,
+                                    FileTypes::Config(cf.clone()),
+                                )
+                            }
+                        }
+                    }
+                }
+            } else {
+                // Case where profile_name is at the beginning or at the middle of existing_profile_line_numbers
+                #[allow(clippy::needless_range_loop)]
```
kvcache, line 95: Hmm here three

src/utils/ini_config.rs
```diff
@@ -0,0 +1,179 @@
...
+                updated_file_contents[index]
+                    .replace_range(AFTER_TOKEN_INDEX..line_len, &format!("{}\n", &cr.token));
+            }
+            updated_file_contents
+        }
+        FileTypes::Config(cf) => {
+            // Check if line is not a comment or profile and for cache
+            if !updated_file_contents[index].starts_with('#')
+                && !updated_file_contents[index].starts_with('[')
+                && updated_file_contents[index].starts_with('c')
```
kvcache, line 157: Another example of brittleness in the config. Hopefully we don't name a new configuration option coalesce_micros or something :-)

Re::new(r#"^cache\s*=\s*(\w+)$"#).replace(0, cf.cache) or something like that would account for whitespace, avoid hard coded indices and account for future configurations. Also note that it would not need to be guarded with an if condition - the value is only replaced if the line exactly matches the grammar of a cache setting line.

tests/configure_profiles_test.rs
```diff
@@ -1,14 +1,45 @@
 #[cfg(test)]
 mod tests {
 
+    use assert_cmd::Command;
+
+    async fn configure_momento_default_profile() {
+        let test_auth_token = std::env::var("TEST_AUTH_TOKEN_DEFAULT").unwrap();
+        let mut cmd = Command::cargo_bin("momento").unwrap();
+        #[allow(clippy::useless_format)]
```
kvcache, line 9: The format on the following expression is useless - address the linter rule instead of disabling it

kvcache (commented):
This is going to be very good for users. I have a couple comments and a general concern around manual parsing without either an ini library or even regular expressions to make it a little more resilient to vimmers or $future_developer's additions.

src/utils/ini_config.rs
poppoerika: Hmm it's interesting... I removed it to type the error clippy gave me (which I didn't believe work for our case at the time), but now it does not give any errors. 
So just removed all three of them.

kvcache (approved):
This looks great, thanks for improving `momento`!

PR #82, opened 2022-03-21
fix: set github origin url with credentials

bruuuuuuuce:


.github/workflows/cd.yml
```diff
@@ -114,3 +96,42 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+          token: ${{ secrets.PRIVATE_REPO_RELEASE_ACCESS }}
+          
+      - name: Update Cargo Version
+        run: |
+          chmod +x set_cargo_version.sh
+          ./set_cargo_version.sh ${{ needs.release.outputs.version }}
```
poppoerika, line 116: Do we need to call `./set_cargo_verions.sh` again here in addition to the `release` step?

bruuuuuuuce: Yea because we are checking out from main again so the cargo version is not the correct version still

poppoerika (approved):

PR #85, opened 2022-03-28
chore: Set file permission 600 for both credentials and config files

### LINKED ISSUE #67: Cred and config file permissions should match how aws cli sets them
You get terminal prompt asking if your sure ok to remove credential file because we dont give user write access. Also we should probably only give r/w to local user vs other users on system.

Screen Shot 2022-03-14 at 3 04 45 PM
Screen Shot 2022-03-14 at 3 05 43 PM

Comparison of aws vs momento
Screen Shot 2022-03-14 at 3 06 41 PM

poppoerika:
Closes #67


kvcache (commented):
Looks great other than the failing test

eaddingtonwhite (approved):

PR #91, opened 2022-04-01
feat: support create/revoke signing key

wandaitzuchen:
Closes https://github.com/momentohq/momento-cli/issues/90

CLI command help:

```
% ./target/release/momento -h
momento 0.11.9
CLI for Momento APIs

USAGE:
    momento [OPTIONS] <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
    -V, --version    Print version information
        --verbose

SUBCOMMANDS:
    account      Account Managements
    cache        Cache Operations
    configure    Configure Momento Credentials
    help         Print this message or the help of the given subcommand(s)
    
% ./target/release/momento account -h
momento-account
Account Managements

USAGE:
    momento account [OPTIONS] <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
        --verbose

SUBCOMMANDS:
    create-signing-key    Creates a Momento Signing Key
    help                  Print this message or the help of the given subcommand(s)
    revoke-signing-key    Revokes the given Momento Signing Key
    signup                Sign up for Momento

% ./target/release/momento account create-signing-key -h
momento-account-create-signing-key
Creates a Momento Signing Key

USAGE:
    momento account create-signing-key [OPTIONS]

OPTIONS:
    -h, --help                 Print help information
    -p, --profile <PROFILE>    [default: default]
    -t, --ttl <TTL_MINUTES>    Duration, in minutes, that the signing key will be valid [default:
                               86400]
        --verbose

% ./target/release/momento account revoke-signing-key -h
momento-account-revoke-signing-key
Revokes the given Momento Signing Key

USAGE:
    momento account revoke-signing-key [OPTIONS] --key-id <KEY_ID>

OPTIONS:
    -h, --help                 Print help information
    -k, --key-id <KEY_ID>      Signing Key ID
    -p, --profile <PROFILE>    [default: default]
        --verbose
```
    
Example: 
```
% ./target/release/momento account create-signing-key
{
  "key_id": "7uqm3ttvl",
  "key": "{\"p\":\"vj3UqtpdAF0AbF-8tdpPofrV5mjKJ41L_EP_FIHkNA4focyTxdjm2jA7Hfa_3nBwWmQ-en7oOrGcBrNNh4qVwWngviOmLmskRqjUQFRGL3PJx3yFYpEKoBE8H2DF2X5j_oih1tDEq-BwiGxMCpDLudOUxUxHI1agokbW9cjWHF0\",\"kty\":\"RSA\",\"q\":\"1NKBO6ELvK4cpkXZONfmvPthS1s9WRp39ZIXm198HWfvy0vjccYztanCm9WgN-n1y6tBOsDJ0WyA3hHnkOshFrwMJohzfWPyXRQATACrPRNxaCwZLGkz2tmz5StELOqQm6zmbZM182LFIn1hzPCvTaGQz4oyEkxNk7eJIJCnq7M\",\"d\":\"q4zRMYhTWPyRQxgCDXQWJQ-1_Zq41GxxW5d133GRI24wDcu283hifwM5-_MM_KIpnRl4d4NsMzX83KTfIf0Qa_sH7Ye54jAkhYoLU90fWdado0SSf8EJeqW9AM05weOGkY5CJBXOWVtzEf8kH-b0zzCQccGmzzzku8qYGbxK1ZQendvxcrRU_6--SthGRXAgQDFTINPcrDxVT6JTVAFM6GglCIGgCLMdHjw0w9IZp3s8K8nT4efGGZDhYk-_mXwKhSXqukw-xXS7-k0pPa69QeTx18VKRpxpHvGKosFRRox9lIQGWvAzfobTWCb5k0X-aAhc6PIZwiNn35nLbkYF\",\"e\":\"AQAB\",\"kid\":\"7uqm3ttvl\",\"qi\":\"RA6giRplw67k_i5AxbSeol-Z_l0r2ogDSH1rFTILc5hmaSOAQ2Y3hID3gMFeaufAuNYmUTx1VqyrXPGRiNTW9rZUjz9MtOmAMUjddgS3fXI_uI4lAeVhNbvZkIHjF8msI_O41WcIN_Kae0nxS1bVT1jI_Vky2WooWYuLlNXNPjQ\",\"dp\":\"mbKyouaQxL5enTAr8nBSbfeN4P1Adfv-K-1zpEnc6QGQJNR3fxziRd05pde45CnLW_wpTiZB0Um1WOR7-snIe97iqDt9A3ukb0HvWZm8qpHNr9G2og-0meXSwTcaIcGvX_Fbb4ZznrlRsaZtOQwKzDNYup4Jy08tsG2zNhfwhDE\",\"alg\":\"RS256\",\"dq\":\"xscG5cb5y3VkVPkQdsNRRFoW-edyWXMbcEbPvdNlfaHCm0N6YJ4bViYFGsBhMVhYt65w8ot29SA_DSvz-atj8HQbdTIljIx5pu4ziMlnJ1itp_roWE66nEKSgA31U26TOpQ3dUj2Lymlrx0k5-I2GnA5bV9KfUS_gKUAc2Yo9Xk\",\"n\":\"niei32kT257jbEHNawDlfZhOe8B3zhfmsriU51VpNZjTiZzMg9SWOhCroXO7Hj-W1gABAGJ68FcT6Z70Xmi5qxpyHvCqkIlgFV2EXey19c-GwpYSA6HIuoxHyGvp5eYafjdAwFfD8rcgFtHV439wP8hg92PbU3pO-XvLRBBRNdGE75UQD3SIYZMJBvoQP_BQ64hXm4tnzq4KKYvFiOuVIOCDSfOGa3zf45fgvVYPzyApBaf2lbVuK3rwslgPNtDJiYYd6AG8-Izxepvvp1JNxggMDu8YeEmfZ77W7j5I1wF0BRU34ylPYDndfHmB_z9-sKu26FXqoluh2T9DQ1L0Bw\"}",
  "expires_at": 1654217009,
  "endpoint": "cache.developer-wanda-dev.preprod.a.momentohq.com"
}

% ./target/release/momento account revoke-signing-key -k 7uqm3ttvl
```
```
% ./target/release/momento account create-signing-key --ttl 120 --profile external-beta
{
  "key_id": "3bf2dhqse",
  "key": "{\"p\":\"xPdTdy85ERyYHLdgN_zbJjsZLm-syRKbTw95goSbwLfy7WGA2e12jRm1rk9OyqbgXWUr1SZWRb9Af-wDyH-NAX8pIdpaXii4ro_siUB2yiFzpPmvgi5LLMBDKnkPtGESPyy8ploguTEm1O3GjhFhSArn4marrzFnxCdgBWZ9bR0\",\"kty\":\"RSA\",\"q\":\"2RCXL-joQ65sbCP9jG2W0SRC4AUUYhkVC7JO94OczpSFgOnbc-3aHmwFdMLvzmPZAvZKEHBfI5EcVhTgiVrl5EApHIAmD4FDCtRWrGT4zZgZGRvXlBvPlRptMw1t5RYq-ooCFkGrbYXjJM3xC9KR5FMFgLZ2bW6zxAWY9mfytfs\",\"d\":\"Du6Qx-dWp1_83BTrDZnUbpny-xeMav6zy9jfnqIkZlqIu5w9CA9YDZVNv2NsfR8gfjveRhP9kgU0Rso2X9Cug9VqNPImtvvFwm5_xtAxWfKZGxd-Skctnoyna6uWaxjL21cXpFS8mMY2aWUbIGPsD9Faz1gNzfdBKlbG_Od0kemoh3Lrc1n0j8tNUrd3_OYR2hDlYpt1oYMiovqC1Phk3IW6Xoih6PVPqJfFRq1Nreydd1BeMjfKiHcO_pz6vc6xnLdU6SS273-FbrJEYk7BbrT-pcpdX458XQha6PRQhgGIoxcBUbt9XhS1YEhpvFMqUFnUxz2LEv89Hghqy_91tQ\",\"e\":\"AQAB\",\"kid\":\"3bf2dhqse\",\"qi\":\"SNGfV7SenbpV_5OAhQTlUpUGU1amR7hcMWJkllfiDXQnhFmcINVSjTTcju8ZMPGlbjI__ZYLCC4ESOdvUXFguzmYRGEUpa4ZymDVuRR1r4bE6SMZZbeV1vPlxV0p2V4hq0Ur3b2Q78eqW_U_EirQfkSBtnR6lKMjVu6DKvFNHbI\",\"dp\":\"pshe7r8X7Xtu-nH0kbQUxaypR_7cKm8M_8c0tuYfFPMuSv-TIfXHuvcYNn-FUmg3-B2hHh4WU3z82tU45t_KwIeyhq3kc3K732STgC0L6WCmEf4zNWKJ2Vz8Asz8FVtVHhV9sMcEiowqO4cT6FyjKWPCsp0Jp5GIMjq9PhVpA6k\",\"alg\":\"RS256\",\"dq\":\"AdQJ8fVcP3WwpBiJgi1KH-jWR1zDly_u5XJTB_Kr13GYzkUp_Af81lTClwEtU-ezZDr-1LSdmIcYXgE_AukpFosTzZGmwGquTeEL-de_4LtvOq7pWe4CCZd3Y_6w9kUBh6qx6c4Ty5BdIWo6pYbYg7vPSA5BYRqVmQkTCMBHYxE\",\"n\":\"pwJph_1Qyqk4eNYKugaZPR3jKqR4bsI8Rauy_0TKetpHFun8j8Fw9OWEQdkRLAtXreEL0NOeIq4HeH7tBJyZMt-mRuyQqZ_1b53hx8UdCbgAbGBwn0UA4jzFZ4XvS_bHiXev72oY7OvNNyMgotVxsPCVRlALzA40dB_zlASyDUEYkD-Jl_vt08WEbFdw8NZaupg3ENnoTWKjEYX4_mD3JUNHDPMpjfNENryjeR2P5hpoEL5sF1_Vi2nPXsIVHYJ9vXCt06il8hdgcuhF27-TxO2CSQnV8_E0bm9wnASDLRmlet_fFxTgjWftrK2_-fcW8BNnIuoSVpdlt2IX-4l8bw\"}",
  "expires_at": 1649042042,
  "endpoint": "cache.cell-external-beta-1.prod.a.momentohq.com"
}

% ./target/release/momento account revoke-signing-key  --profile external-beta --key-id 3bf2dhqse
```

### DIFF
--- a/src/commands/signingkey/signingkey_cli.rs
+++ b/src/commands/signingkey/signingkey_cli.rs
@@ -1,4 +1,4 @@
-use log::info;
+use log::debug;
 use momento::simple_cache_client::SimpleCacheClient;
 
 use crate::error::CliError;
@@ -11,13 +11,11 @@ async fn get_momento_instance(auth_token: String) -> Result<SimpleCacheClient, C
 }
 
 pub async fn create_signing_key(ttl_minutes: u32, auth_token: String) -> Result<(), CliError> {
-    info!("creating signing key...");
+    debug!("creating signing key...");
     let mut momento = get_momento_instance(auth_token).await?;
     match momento.create_signing_key(ttl_minutes).await {
         Ok(res) => {
-            println!("endpoint = {}", res.endpoint);
-            println!("key = {}", res.key);
-            println!("expires_at = {}", res.expires_at);
+            println!("{}", serde_json::to_string_pretty(&res).unwrap());
         }
         Err(e) => return Err(CliError { msg: e.to_string() }),
     };
@@ -25,7 +23,7 @@ pub async fn create_signing_key(ttl_minutes: u32, auth_token: String) -> Result<
 }
 
 pub async fn revoke_signing_key(key_id: String, auth_token: String) -> Result<(), CliError> {
-    info!("revoking signing key...");
+    debug!("revoking signing key...");
     let mut momento = get_momento_instance(auth_token).await?;
     match momento.revoke_signing_key(&key_id).await {
         Ok(_) => (),
--- a/src/commands/signingkey/signingkey_cli.rs
+++ b/src/commands/signingkey/signingkey_cli.rs
@@ -1,0 +1,36 @@
+use log::info;
+use momento::simple_cache_client::SimpleCacheClient;
+
+use crate::error::CliError;
+
+async fn get_momento_instance(auth_token: String) -> Result<SimpleCacheClient, CliError> {
+    match SimpleCacheClient::new(auth_token, 100).await {
+        Ok(m) => Ok(m),
+        Err(e) => Err(CliError { msg: e.to_string() }),
+    }
+}
+
+pub async fn create_signing_key(ttl_minutes: u32, auth_token: String) -> Result<(), CliError> {
+    info!("creating signing key...");
+    let mut momento = get_momento_instance(auth_token).await?;
+    match momento.create_signing_key(ttl_minutes).await {
+        Ok(res) => {
+            println!("user_id = {}", res.user_id);
+            println!("endpoint = {}", res.endpoint);
+            println!("key = {}", res.key);
+            println!("expires_at = {}", res.expires_at);
+        }
+        Err(e) => return Err(CliError { msg: e.to_string() }),
+    };
+    Ok(())
+}
+
+pub async fn revoke_signing_key(key_id: String, auth_token: String) -> Result<(), CliError> {
+    info!("revoking signing key...");
+    let mut momento = get_momento_instance(auth_token).await?;
+    match momento.revoke_signing_key(&key_id).await {
+        Ok(_) => (),
+        Err(e) => return Err(CliError { msg: e.to_string() }),
+    };
+    Ok(())
+}
--- a/src/commands/signingkey/signingkey_cli.rs
+++ b/src/commands/signingkey/signingkey_cli.rs
@@ -15,7 +15,6 @@ pub async fn create_signing_key(ttl_minutes: u32, auth_token: String) -> Result<
     let mut momento = get_momento_instance(auth_token).await?;
     match momento.create_signing_key(ttl_minutes).await {
         Ok(res) => {
-            println!("user_id = {}", res.user_id);
             println!("endpoint = {}", res.endpoint);
             println!("key = {}", res.key);
             println!("expires_at = {}", res.expires_at);
--- a/src/main.rs
+++ b/src/main.rs
@@ -56,7 +56,7 @@ enum AccountCommand {
         region: String,
     },
 
-    #[structopt(about = "Creates a Momento Signing Key")]
+    #[structopt(about = "Create a Momento Signing Key")]
     CreateSigningKey {
         #[structopt(
             long = "ttl",
@@ -69,7 +69,7 @@ enum AccountCommand {
         profile: String,
     },
 
-    #[structopt(about = "Revokes the given Momento Signing Key")]
+    #[structopt(about = "Revoke the given Momento Signing Key")]
     RevokeSigningKey {
         #[structopt(long = "key-id", short, help = "Signing Key ID")]
         key_id: String,
@@ -80,7 +80,7 @@ enum AccountCommand {
 
 #[derive(Debug, StructOpt)]
 enum CacheCommand {
-    #[structopt(about = "Creates a Momento Cache")]
+    #[structopt(about = "Create a Momento Cache")]
     Create {
         #[structopt(long = "name", short = 'n')]
         cache_name: String,
@@ -88,7 +88,7 @@ enum CacheCommand {
         profile: String,
     },
 
-    #[structopt(about = "Stores a given item in cache")]
+    #[structopt(about = "Store a given item in cache")]
     Set {
         #[structopt(long = "name", short = 'n')]
         cache_name: Option<String>,
@@ -107,7 +107,7 @@ enum CacheCommand {
         profile: String,
     },
 
-    #[structopt(about = "Gets item from the cache")]
+    #[structopt(about = "Get item from the cache")]
     Get {
         #[structopt(long = "name", short = 'n')]
         cache_name: Option<String>,
@@ -118,7 +118,7 @@ enum CacheCommand {
         profile: String,
     },
 
-    #[structopt(about = "Deletes the cache")]
+    #[structopt(about = "Delete the cache")]
     Delete {
         #[structopt(long = "name", short = 'n')]
         cache_name: String,
@@ -126,7 +126,7 @@ enum CacheCommand {
         profile: String,
     },
 
-    #[structopt(about = "Lists all momento caches")]
+    #[structopt(about = "List all momento caches")]
     List {
         #[structopt(long, short, default_value = "default")]
         profile: String,
--- a/src/main.rs
+++ b/src/main.rs
@@ -24,17 +24,17 @@ struct Momento {
 
 #[derive(Debug, StructOpt)]
 enum Subcommand {
-    #[structopt(about = "Interact with Momento Caches")]
+    #[structopt(about = "Interact with caches")]
     Cache {
         #[structopt(subcommand)]
         operation: CacheCommand,
     },
-    #[structopt(about = "Configure Momento Credentials")]
+    #[structopt(about = "Configure credentials")]
     Configure {
         #[structopt(long, short, default_value = "default")]
         profile: String,
     },
-    #[structopt(about = "Manage Accounts")]
+    #[structopt(about = "Manage accounts")]
     Account {
         #[structopt(subcommand)]
         operation: AccountCommand,
@@ -56,7 +56,7 @@ enum AccountCommand {
         region: String,
     },
 
-    #[structopt(about = "Create a Momento Signing Key")]
+    #[structopt(about = "Create a signing key")]
     CreateSigningKey {
         #[structopt(
             long = "ttl",
@@ -69,7 +69,7 @@ enum AccountCommand {
         profile: String,
     },
 
-    #[structopt(about = "Revoke the given Momento Signing Key")]
+    #[structopt(about = "Revoke the signing key")]
     RevokeSigningKey {
         #[structopt(long = "key-id", short, help = "Signing Key ID")]
         key_id: String,
@@ -80,7 +80,7 @@ enum AccountCommand {
 
 #[derive(Debug, StructOpt)]
 enum CacheCommand {
-    #[structopt(about = "Create a Momento Cache")]
+    #[structopt(about = "Create a cache")]
     Create {
         #[structopt(long = "name", short = 'n')]
         cache_name: String,
@@ -118,7 +118,7 @@ enum CacheCommand {
         profile: String,
     },
 
-    #[structopt(about = "Delete the given Momento Cache")]
+    #[structopt(about = "Delete the cache")]
     Delete {
         #[structopt(long = "name", short = 'n')]
         cache_name: String,
@@ -126,7 +126,7 @@ enum CacheCommand {
         profile: String,
     },
 
-    #[structopt(about = "List all Momento Caches")]
+    #[structopt(about = "List all caches")]
     List {
         #[structopt(long, short, default_value = "default")]
         profile: String,
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,7 +3,7 @@ use std::{panic, process::exit};
 use clap::StructOpt;
 use env_logger::Env;
 use error::CliError;
-use log::{error, info};
+use log::{error, debug};
 use utils::user::get_creds_and_config;
 
 pub mod commands;
@@ -29,17 +29,12 @@ enum Subcommand {
         #[structopt(subcommand)]
         operation: CacheCommand,
     },
-    #[structopt(about = "Signing Key Operations")]
-    SigningKey {
-        #[structopt(subcommand)]
-        operation: SigningKeyCommand,
-    },
     #[structopt(about = "Configure Momento Credentials")]
     Configure {
         #[structopt(long, short, default_value = "default")]
         profile: String,
     },
-    #[structopt(about = "Manage Accounts")]
+    #[structopt(about = "Account Managements")]
     Account {
         #[structopt(subcommand)]
         operation: AccountCommand,
@@ -59,6 +54,27 @@ enum AccountCommand {
             help = "e.g. us-west-2, us-east-1, ap-northeast-1"
         )]
         region: String,
+    },
+
+    #[structopt(about = "Creates a Momento Signing Key")]
+    CreateSigningKey {
+        #[structopt(
+            long = "ttl",
+            short = 't',
+            default_value = "86400",
+            help = "Duration, in minutes, that the signing key will be valid"
+        )]
+        ttl_minutes: u32,
+        #[structopt(long, short, default_value = "default")]
+        profile: String,
+    },
+
+    #[structopt(about = "Revokes the given Momento Signing Key")]
+    RevokeSigningKey {
+        #[structopt(long = "key-id", short, help = "Signing Key ID")]
+        key_id: String,
+        #[structopt(long, short, default_value = "default")]
+        profile: String,
     },
 }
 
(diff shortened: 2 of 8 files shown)


virratanasangpunth:
```
./target/release/momento signingkey revoke -k le6lb4ux3
```

Are we assuming that the customers are smart enough to determine where `le6lb4ux3` is coming from?

danielamiao:
> % ./target/release/momento signingkey -h
Should we change to `momento signing-key -h`? Finding the hyphenated version more readable.

> Revokes the Signing Key
Revokes the given Momento Signing Key

> -t, --ttl <TTL_MINUTES>    Timespan, in minutes, that the signing key will be valid
I'm torn, should we make this optional/default to 90 days? If not, we should at least say `Duration, in minutes, that the signing key will be valid. Recommended value is 129600 minutes (90 days).` Timespan is a technical word :P

wandaitzuchen:
Yeah I think making it optional makes sense.  The max allowed ttl is 60 days btw.

danielamiao:
> Yeah I think making it optional makes sense. The max allowed ttl is 60 days btw.

OK perfect, sorry I misremembered. If you make it optional and default to 60 days. I suggest the following copy: `Duration, in minutes, that the signing key will be valid. Default value is 86400 minutes (60 days).`

danielamiao:
Wording/copy all looks good to me. Someone who's familiar with the CLI should take a look at this as well @kvcache @bruuuuuuuce @poppoerika 

Looking at this PR made me realize we need to think about adding a `ListSigningKeys` API soon

src/commands/signingkey/signingkey_cli.rs
```diff
@@ -0,0 +1,35 @@
...
+
+pub async fn create_signing_key(ttl_minutes: u32, auth_token: String) -> Result<(), CliError> {
+    info!("creating signing key...");
+    let mut momento = get_momento_instance(auth_token).await?;
+    match momento.create_signing_key(ttl_minutes).await {
+        Ok(res) => {
+            println!("endpoint = {}", res.endpoint);
+            println!("key = {}", res.key);
+            println!("expires_at = {}", res.expires_at);
+        }
```
kvcache, line 21: if you pretty-printed a json it would be friendlier to user scripts

How is a user supposed to use the key id? Aren't we telling them to give the key id back to us when they ask us to revoke a key? Are we going to teach users what the contents of JWS's are and how to base64 decode them or should we let users treat them as opaque strings and explicitly tell them what they need to know? (i'm strongly biased for the latter)

src/commands/signingkey/signingkey_cli.rs
```diff
@@ -0,0 +1,35 @@
...
+
+async fn get_momento_instance(auth_token: String) -> Result<SimpleCacheClient, CliError> {
+    match SimpleCacheClient::new(auth_token, 100).await {
+        Ok(m) => Ok(m),
+        Err(e) => Err(CliError { msg: e.to_string() }),
+    }
+}
+
+pub async fn create_signing_key(ttl_minutes: u32, auth_token: String) -> Result<(), CliError> {
+    info!("creating signing key...");
```
kvcache, line 14: if you made debug output happen at the `debug` level instead it would be friendlier to user scripts

src/commands/signingkey/signingkey_cli.rs
```diff
@@ -0,0 +1,35 @@
...
+            println!("key = {}", res.key);
+            println!("expires_at = {}", res.expires_at);
+        }
+        Err(e) => return Err(CliError { msg: e.to_string() }),
+    };
+    Ok(())
+}
+
+pub async fn revoke_signing_key(key_id: String, auth_token: String) -> Result<(), CliError> {
+    info!("revoking signing key...");
```
kvcache, line 28: 🤔 how many tools say what they're going to do by default when their job is to do 1 function? I think the `momento` application should do what it was asked to do, and by default only output information that is valuable. For example, if I write `momento signing-key revoke --key id123` and press enter, isn't it my default expectation that the `momento` program will revoke a signing key? I think this is a debug level output.

src/main.rs
```diff
@@ -29,9 +29,14 @@ enum Subcommand {
         #[structopt(subcommand)]
         operation: CacheCommand,
     },
+    #[structopt(about = "Signing Key Operations")]
+    SigningKey {
+        #[structopt(subcommand)]
+        operation: SigningKeyCommand,
+    },
```
kvcache, line 36: Should this be at the root level of `momento` or is it better grouped under `account`? This addition would mean that 3 out of 4 of our root-level subcommands have nothing to do with caching, and are only concerned with configuration & setup. Can we come up with a command tree that makes sense contextually?

src/main.rs
```diff
@@ -83,32 +87,55 @@ enum CacheCommand {
...
+            help = "Duration, in minutes, that the signing key will be valid. Default value is 86400 minutes (60 days)."
+        )]
+        ttl_minutes: Option<u32>,
+        #[structopt(long, short, default_value = "default")]
+        profile: String,
+    },
+
+    #[structopt(about = "Revokes the given Momento Signing Key")]
+    Revoke {
+        #[structopt(long = "kid", short, help = "Signing Key ID")]
```
kvcache, line 136: why `kid`? We don't spare characters in `long`s when it makes the resulting argument obscure.

When you printed it for the user you called the key `key` and you never showed them a `kid`. `kid` is a jws compact header identifier right? Why not let it be called `--key-id` in the CLI?

src/main.rs
```diff
@@ -83,32 +87,55 @@ enum CacheCommand {
...
+        key_id: String,
```
kvcache, line 137: I think we should also (or instead) provide `--key` so if a user doesn't care to remember their `key-id` they can just give us the JWS token and we can revoke it.

src/main.rs
```diff
@@ -178,6 +205,29 @@ async fn entrypoint() -> Result<(), CliError> {
...
+                .await?;
+            }
+            SigningKeyCommand::Revoke { key_id, profile } => {
+                let (creds, _config) = get_creds_and_config(&profile).await?;
+                commands::signingkey::signingkey_cli::revoke_signing_key(
+                    key_id.clone(),
+                    creds.token,
+                )
+                .await?;
+                info!("revoked signing key {}", key_id)
```
kvcache, line 228: Again, this is a debug message. If `$?` is `0` then it worked. If that's ever untrue then we have a bug.

src/main.rs
```diff
@@ -83,32 +87,55 @@ enum CacheCommand {
...
+#[derive(Debug, StructOpt)]
+enum SigningKeyCommand {
+    #[structopt(about = "Creates a Momento Signing Key")]
+    Create {
+        #[structopt(
+            long = "ttl",
+            short = 't',
+            help = "Duration, in minutes, that the signing key will be valid. Default value is 86400 minutes (60 days)."
+        )]
+        ttl_minutes: Option<u32>,
```
kvcache, line 129: remove `Option` and put the default here instead of using unwrap below. Clap supports this and you've got a handy example 1 line below.

kvcache (commented):
I was pinged so I looked through this & have some notes

mjindalmotiv8:
@danielamiao - created https://github.com/momentohq/cacheadmin-v2/issues/532 to track that we need to support List

kvcache:
> ```
> ./target/release/momento signingkey revoke -k le6lb4ux3
> ```
> 
> Are we assuming that the customers are smart enough to determine where `le6lb4ux3` is coming from?

I echo this concern and am strongly opposed to making this assumption. I don't want to teach all our users about JWS encoding and structure; they need to be able to use our auth tokens end-to-end without caring what's inside. They can if they want, but there should never be a functional reason for them to _need_ to.

src/main.rs
danielamiao: Yeah this is a good idea. I would probably make it an additional though, so one can revoke using key id OR the actual key

src/main.rs
danielamiao: Good call

src/main.rs
kvcache: yeah - if we print a `key_id` to the user then I agree we need to support revocation via that `key_id`. Currently we don't print the key id BUT probably as an outcome from this feedback we will.

src/commands/signingkey/signingkey_cli.rs
danielamiao: Per comment below I think it makes sense to parse out key id and return it as a separate 4th field in the response

wandaitzuchen:
PR updated:
- `create-signing-key`command now under `account`
- `create-signing-key` response now includes the key_id.
- moved a few output messages to debug level

kvcache (commented):
Looks good. One doc nit & it's ready to go

src/main.rs
```diff
@@ -31,10 +31,10 @@ enum Subcommand {
     },
     #[structopt(about = "Configure Momento Credentials")]
     Configure {
-        #[structopt(name = "profile", long, short, default_value = "default")]
+        #[structopt(long, short, default_value = "default")]
         profile: String,
     },
-    #[structopt(about = "Manage Accounts")]
+    #[structopt(about = "Account Managements")]
```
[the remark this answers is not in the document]
wandaitzuchen: I changed it thinking this would be more consistent with what we have, i.e. print what a leaf subcommands does, but for subcommand that have children subcommands, print what it is. E.g. for `cache`, we print `Cache Operations`. 

If we'd like to use `Manage Accounts`, we should probably update the about for `cache` as well?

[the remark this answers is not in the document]
kvcache: Oh yeah, was on mobile (computer was updating) and I didn't see Cache. That should really be `Interact with Caches` or something

src/main.rs
```diff
@@ -80,15 +80,15 @@ enum AccountCommand {
...
+    #[structopt(about = "Create a Momento Cache")]
     Create {
         #[structopt(long = "name", short = 'n')]
         cache_name: String,
         #[structopt(long, short, default_value = "default")]
         profile: String,
     },
 
-    #[structopt(about = "Stores a given item in cache")]
+    #[structopt(about = "Store a given item in cache")]
```
kvcache, line 91: while you're changing the voice, how about making get and set docs consistent?
```
Store an item in a Cache
```
```
Retrieve an item from a Cache
```

kvcache: oh looks like we use cache and Cache inconsistently. `cache` is fine then

src/main.rs
```diff
@@ -118,15 +118,15 @@ enum CacheCommand {
...
+    #[structopt(about = "Delete the cache")]
     Delete {
         #[structopt(long = "name", short = 'n')]
         cache_name: String,
         #[structopt(long, short, default_value = "default")]
         profile: String,
     },
 
-    #[structopt(about = "Lists all momento caches")]
+    #[structopt(about = "List all momento caches")]
```
kvcache, line 129: elsewhere we capitalize Momento

src/main.rs
```diff
@@ -69,7 +69,7 @@ enum AccountCommand {
         profile: String,
     },
 
-    #[structopt(about = "Revokes the given Momento Signing Key")]
+    #[structopt(about = "Revoke the given Momento Signing Key")]
```
kvcache, line 72: Is "Signing Key" a Momento-specific noun? I think it's just `signing key` right?

wandaitzuchen: I think it's fine to use `Momento Signing Key`, like `Momento Cache`.  These are all entities that Momento creates for customers.
@danielamiao any preferences?

danielamiao: Very slight preference for `Momento signing key` just to keep it consistent with `Momento cache` which is mentioned elsewhere.

wandaitzuchen: it's `Momento Cache` elsewhere in the doc currently.  I'm going to update them all to lowercase.

src/main.rs
```diff
@@ -118,15 +118,15 @@ enum CacheCommand {
         profile: String,
     },
 
-    #[structopt(about = "Delete the cache")]
+    #[structopt(about = "Delete the given Momento Cache")]
```
kvcache, line 121: we're not using `Momento Cache` all over the place in these docs - just leave it `cache` (here and everywhere). These docs are for helping users use the `momento` tool, they're not an advertisement space for Momento® Cache® or whatever 😄

Each change you make in these docs should (a) be consistent with the existing docs or (b) redo the existing docs to be consistent with your proposal.

kvcache: This is very much bikeshedding - please just make it consistent so we can move on 😂

kvcache (approved):
lgtm

PR #92, opened 2022-04-04
Update README.md

ksmotiv8:
Made our getting started punchy and to the point

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,28 @@
 # Momento CLI
 
-_Read this in other languages_: [日本語](README.ja.md)
-<br >
+Japanese: [日本語](README.ja.md)
 
-## Prerequisites
 
-- MacOS or Linux
-- [Homebrew](https://brew.sh/)
 
-## Install
 
+
+
+## Quick Start
 ```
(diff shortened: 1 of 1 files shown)


eaddingtonwhite (approved):
Looks good to me. Maybe only concern is no `configure` step up above after `signup` cmd so user might not know how to set auth token after obtaining? Maybe we add to quickstart block. Will let @poppoerika take final pass and merge if she is good.

README.md
```diff
@@ -32,6 +43,10 @@ momento account signup --email <TYPE_YOUR_EMAIL_HERE>
 # (optional) view help to see all available regions, and sign up for a specific region
 momento account signup --help
 momento account signup --email <TYPE_YOUR_EMAIL_HERE> --region <TYPE_DESIRED_REGION>
+
+# Configure CLI
+momento configure
```
danielamiao, line 48: We cannot skip this step in `Quickstart` above, without `momento configure` the CLI does not have the auth token and create cache, set/get will all fail.

poppoerika: +

PR #103, opened 2022-04-11
chore: Use publicly available momento crate and remove client-sdk-rust submodule

### LINKED ISSUE #101: Consume Momento Rust SDK from crates
Once we have a crate for our Rust SDK, we can consume that instead of using git submodules.

poppoerika:
Closes #101


tylerburdsall (approved):
Nice, this is so great to not have to use submodules now!

bruuuuuuuce (approved):
:shipit:  :celebrate:

PR #112, opened 2022-04-12
chore: Add `--quick` flag to use default cache name and TTL on `momento configure`

### LINKED ISSUE #109: Explore better messaging for defaults with configure
Feedback from user:
 
`momento configure` includes these prompts:

    Default Cache [default-cache]:
    Default Ttl Seconds [600]:

I had to stop and think for a few seconds at each of these prompts before deciding nothing mattered at this stage while I was testing it out.

If the goal is to get folks up and running quickly, consider having the copy-paste fast-start demo instructions be something like:

    momento configure --defaults

or

    momento configure defaults

and then only ask for the token.

Alternatively, the default could be to default the defaults when they type:

    momento configure

and then support a path that prompts for more configuration option details when they type something like:

    momento configure --full

This alternative approach would keep the fast-start instructions super clean, while not losing functionality.

Upon reading the documentation further after a fast-start trial, developers can discover that they can change the default cache name and TTL, but it doesn't seem important for getting up and learning about the service.

poppoerika:
Added `--quick` flag to use the default cache name and TTL without users entering anything.
Users are still required to enter their auth token.

First time creating `default-cache`
```
./target/debug/momento configure --quick
Token [****]: gregger
[2022-04-13T18:28:40Z INFO  momento::commands::configure::configure_cli] default-cache successfully created with default TTL of 600s
```

`default-cache` already exists
```
./target/debug/momento configure --quick
Token [****]: gdfgfdg
[2022-04-13T18:29:14Z INFO  momento::commands::configure::configure_cli] default-cache already exists
```

Closes #109

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,22 +2,18 @@
 
 Japanese: [日本語](README.ja.md)
 
-
-
-
-
-
 ## Quick Start
+
 ```
-# Install 
+# Install
 brew tap momentohq/tap
 brew install momento-cli
 
 # Sign Up [available regions are us-west-2, us-east-1, ap-northeast-1, default is us-west-2]
 momento account signup --email <TYPE_YOUR_EMAIL_HERE> --region <TYPE_DESIRED_REGION>
 
-# Configure your account with the credentials in your email
-momento configure
+# Configure your account with the credentials in your email and default cache name and TTL
+momento configure --quick
 
 # Make a cache
 momento cache create --name example-cache
@@ -29,13 +25,13 @@ momento cache get --key key --name example-cache
 ```
 
 ## Upgrading
+
 ```
 brew update momento-cli
 brew upgrade momento-cli
 ```
 
-## Sign up 
-
+## Sign up
 
 **NOTE:** If you run into errors during signup, please ensure you have upgraded to the [latest version](https://github.com/momentohq/momento-cli/releases/latest) of our CLI.
 
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ brew install momento-cli
 # Sign Up [available regions are us-west-2, us-east-1, ap-northeast-1, default is us-west-2]
 momento account signup --email <TYPE_YOUR_EMAIL_HERE> --region <TYPE_DESIRED_REGION>
 
-# Configure your account with the credentials in your email and default cache name and TTL
+# Configure your account with the credentials in your email, plus default cache name and TTL
 momento configure --quick
 
 # Make a cache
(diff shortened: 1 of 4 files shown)


poppoerika:
@danielamiao 
Understood. I read his feedback again and now understand that the key is that he had to stop and think so improving messaging does not help that part.

I also like his 2nd idea but for the first bullet point, I think we want to pass a flag to use the default cache name and TTL since I think we want to reserve `momento configure` command to configure an existing default profile. 
So instead, how about `momento configure --quick` something along the line to use the default cache name and TTL and only set an auth token?
For setting all properties (cache name, TTL, auth token), we can do `momento configure --full`.

danielamiao:
@poppoerika `momento configure --quick` sounds good, just remember to update the README to this command.

`momento configure` can be reserved for the full properties, you don't need `--full` flag, right?

poppoerika:
@danielamiao 
> momento configure --quick sounds good, just remember to update the README to this command.

Yup, I'll update README 👍 

>momento configure can be reserved for the full properties, you don't need --full flag, right?

Yup, that's correct. We can keep `momento configure` as is for setting cache name, TTL, and auth token.

README.md
```diff
@@ -2,22 +2,18 @@
...
+# Install
 brew tap momentohq/tap
 brew install momento-cli
 
 # Sign Up [available regions are us-west-2, us-east-1, ap-northeast-1, default is us-west-2]
 momento account signup --email <TYPE_YOUR_EMAIL_HERE> --region <TYPE_DESIRED_REGION>
 
-# Configure your account with the credentials in your email
-momento configure
+# Configure your account with the credentials in your email and default cache name and TTL
```
danielamiao, line 15:
```suggestion
# Configure your account with the credentials in your email, plus default cache name and TTL
```

src/commands/configure/configure_cli.rs
```diff
@@ -60,10 +60,13 @@ pub async fn configure_momento(profile_name: &str) -> Result<(), CliError> {
         }
     }
     match create_cache(config.cache, credentials.token).await {
-        Ok(_) => info!("default cache successfully created"),
+        Ok(_) => info!(
+            "default-cache successfully created with default TTL of {}s",
```
danielamiao, line 64: is it always called `default-cache`? can't the user pass in a different cache name for their default cache?

poppoerika: I can update it to
`default cache of <cache name> successfully created with default TTL of <ttl>s` so it's clear what was set as their default cache?

danielamiao: `<cache name> successfully created as the default, with TTL of <ttl>s`

danielamiao (approved):

PR #114, opened 2022-04-13
chore: Improve cache miss handling

### LINKED ISSUE #108: Explore better exit code for cache miss on cli
Customer Feedback:

On a cache miss, it might be nice to set the exit code to non-success. Right now, it is zero (success):

    $ momento cache set -k name -v bilbo -t 10
    $ echo $?
    0

    $ momento cache get -k name
    bilbo
    $ echo $?
    0

    $ momento cache get -k name
    [2022-04-11T21:50:43Z INFO  momento::commands::cache::cache_cli] cache miss
    $ echo $?
    0

If it were non-zero, then developers could use it to go generate values and re-fill the cache.

    key="some key"
    ttl=10

    if ! value=$(momento cache get -k "$key" 2>/dev/null); then
        value=$(echo "slow command here to generate value for key:" "$key")
        momento cache set -k "$key" -v "$value"  -t $ttl
    fi
    echo "Use value here: $value"

but that code doesn't work today because the cache miss command returns success.

Also in the code sample,

    value=$(momento cache get -k "$key" 2>/dev/null)

I use

    2>/dev/null

so that my stderr doesn't get the "cache miss" warnings spewed out to it.

It would be nice to have a command line option that suppresses a cache miss warning, like:

    value=$(momento cache get -k "$key" --quiet)

or

    value=$(momento cache get -k "$key" --quiet-on-miss)

In addition to being clearer, this could allow more important stderr messages to still be seen. For example, I assume there are some errors output if momento can't access the API or if there is some problem with the command line arguments as in:

    momento cache get -k "$key" --bogus 2>/dev/null

poppoerika:
- Exit with code 1 on cache miss
```
./target/debug/momento cache get --key erika
echo $?                                     
1
```

- Add `--quiet` flag on the set command so that `[2022-04-13T20:55:58Z INFO  momento::commands::cache::cache_cli] cache miss` won't be displayed

Closes #108

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -118,6 +118,8 @@ enum CacheCommand {
         key: String,
         #[structopt(long, short, default_value = "default")]
         profile: String,
+        #[structopt(long, short)]
+        quiet: bool,
     },
 
     #[structopt(about = "Delete the cache")]
@@ -178,12 +180,14 @@ async fn entrypoint() -> Result<(), CliError> {
                 cache_name,
                 key,
                 profile,
+                quiet,
             } => {
                 let (creds, config) = get_creds_and_config(&profile).await?;
                 commands::cache::cache_cli::get(
                     cache_name.unwrap_or(config.cache),
                     creds.token,
                     key,
+                    quiet,
                 )
                 .await?;
             }
--- a/src/main.rs
+++ b/src/main.rs
@@ -118,8 +118,6 @@ enum CacheCommand {
         key: String,
         #[structopt(long, short, default_value = "default")]
         profile: String,
-        #[structopt(long, short)]
-        quiet: bool,
     },
 
     #[structopt(about = "Delete the cache")]
@@ -180,14 +178,12 @@ async fn entrypoint() -> Result<(), CliError> {
                 cache_name,
                 key,
                 profile,
-                quiet,
             } => {
                 let (creds, config) = get_creds_and_config(&profile).await?;
                 commands::cache::cache_cli::get(
                     cache_name.unwrap_or(config.cache),
                     creds.token,
                     key,
-                    quiet,
                 )
                 .await?;
             }
--- a/src/commands/cache/cache_cli.rs
+++ b/src/commands/cache/cache_cli.rs
@@ -1,6 +1,7 @@
 use log::{debug, info};
 use momento::simple_cache_client::SimpleCacheClient;
 use std::num::NonZeroU64;
+use std::process::exit;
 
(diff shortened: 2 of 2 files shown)


src/main.rs
```diff
@@ -178,12 +180,14 @@ async fn entrypoint() -> Result<(), CliError> {
...
                 key,
                 profile,
+                quiet,
             } => {
                 let (creds, config) = get_creds_and_config(&profile).await?;
                 commands::cache::cache_cli::get(
                     cache_name.unwrap_or(config.cache),
                     creds.token,
                     key,
+                    quiet,
```
bruuuuuuuce, line 190: do we want to use this quiet variable in other functions as well

poppoerika: Yeah, I was also thinking if we need it. 
It's not a lot of work to add it so I think it's good to have quiet option for other commands.

poppoerika: I will also update README.

bruuuuuuuce: I am wondering if instead of adding a `quiet`, if we just log when they explicity say they want verbosity. It might be strange to have a normal mode, a verbose mode, and a quiet mode

bruuuuuuuce: so maybe just convert the info logs to debug logs wdyt

poppoerika: Oh I see. It makes sense to me.
It looks like we use `debug` log for other subcommands such as create, delete, and list cache so to align with those commands' logging behavior is a good idea.

danielamiao: I agree with @bruuuuuuuce, instead of adding `quiet` mode let's just convert info logs to debug logs. 

The only command that should output something to stdout is get, it should just print out the value (not a log statement)

eaddingtonwhite: I was thinking that he wanted the --quite-on-miss flag to also make it so CLI would exit 0 on miss if that was passed vs always exiting 0 now.

danielamiao: @eaddingtonwhite re-reading EricH's feedback, I don't think that's what he wanted, he wanted 2 things:
1. Cache miss should exit non-zero, we are doing this here in this PR
2. Cache miss should not spew out anything to stderr, currently the INFO log statement goes to stderr, we are getting rid of that in this PR.

danielamiao (approved):
LGTM but let's wait for @eaddingtonwhite to ack my reply make sure we are aligned before merging.

eaddingtonwhite (approved):
LGTM lets get for feedback

Issue #122, opened 2022-04-16
Add Windows support for file permission


poppoerika:
Got an error [here](https://github.com/momentohq/momento-cli/runs/6044124897?check_suite_focus=true#step:4:508) since OS is not unix.
Need to do something like [this](https://github.com/dtolnay/cargo-llvm-lines/pull/9/files).

Also need to install 32-bit mingw since we got an error [here](https://github.com/momentohq/momento-cli/runs/6044124897?check_suite_focus=true#step:4:365) `choco install winlibs --x86`?

PR #123, opened 2022-04-18
fix: Upload Windows binary file 64

poppoerika:
Addresses some of the issues mentioned in https://github.com/momentohq/momento-cli/issues/122
Comment out uploading Windows binary 32 to make sure 64 binary gets uploaded successfully first.

### DIFF
--- a/src/commands/configure/configure_cli.rs
+++ b/src/commands/configure/configure_cli.rs
@@ -1,4 +1,5 @@
 use log::info;
+use std::os::unix::fs::PermissionsExt;
 use std::path::Path;
 use tokio::fs;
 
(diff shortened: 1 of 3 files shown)


src/commands/configure/configure_cli.rs
```diff
@@ -138,6 +137,48 @@ async fn prompt_user_for_config(quick: bool, profile_name: &str) -> Result<Confi
     })
 }
 
+#[cfg(target_os = "ubuntu")]
```
eaddingtonwhite, line 140: what about osx and other linux versions? Are they supported somehow?

eaddingtonwhite: do we need a defaultt unix handler somewhere still?

eaddingtonwhite (approved):

PR #124, opened 2022-04-18
fix: specify each target os

poppoerika:


eaddingtonwhite (approved):
A little suprised we need `linux` and `ubuntu` targets but if this works LGTM

PR #137, opened 2022-04-19
fix: add mingw32 to PATH

poppoerika:


.github/workflows/cd.yml
```diff
@@ -168,6 +168,13 @@ jobs:
           asset_name: ${{ steps.create_release.outputs.asset_name }}
           asset_content_type: application/zip
 
+      - name: add mingw32 to path
+        shell: bash
+        run: |
+          echo "C:\msys64\mingw32\bin" >> $GITHUB_PATH
```
eaddingtonwhite, line 174: this is kinda neat didnt know could do this and they have env var like this. 👍

poppoerika: Thank you, Internet.

.github/workflows/cd.yml
```diff
@@ -168,6 +168,13 @@ jobs:
...
+
+      - run: i686-w64-mingw32-gcc --version
```
eaddingtonwhite, line 176: should we check exit code explicitly? or I guess that happens for us cause if this fails w/ non 0 exit code the action fails right?

eaddingtonwhite (approved):

.github/workflows/cd.yml
poppoerika: Yeah, the last action failed with exit code 1.

PR #150, opened 2022-05-20
chore: upgrade momento sdk - consistency, less copypasta

kvcache:
This change establishes a common pattern for interacting with the
service. Rather than copy/pasting matches, MomentoError -> CliError
mapping and json dump code around, this names those functionalities
as functions:
* interact_with_momento: Completes a momento future and maps the
  result to a Result< HappyCaseType, CliError >
* print_whatever_this_is_as_json: prints whatever you feed it as json

This change was precipitated by the change in the SDK to simplify
and coalesce client creation behind the SimpleCacheClientBuilder and
triggered by the SDK upgrade.

There are small changes to debug and error messages, but this change
should cause no functional changes.

tired of copypasta.
```
pub async fn delete_cache(cache_name: String, auth_token: String) -> Result<(), CliError> {
    debug!("deleting cache...");
    let mut momento = get_momento_client(auth_token).await?;
    match momento.delete_cache(&cache_name).await {
        Ok(_) => (),
        Err(e) => return Err(CliError { msg: e.to_string() }),
    };
    Ok(())
}
```
becomes
```
pub async fn delete_cache(cache_name: String, auth_token: String) -> Result<(), CliError> {
    let mut client = get_momento_client(auth_token).await?;
    interact_with_momento("deleting cache...", client.delete_cache(&cache_name)).await
}
```

### DIFF
--- a/src/utils/client.rs
+++ b/src/utils/client.rs
@@ -1,0 +1,49 @@
+use std::{future::Future, num::NonZeroU64};
+
+use momento::{
+    response::error::MomentoError,
+    simple_cache_client::{SimpleCacheClient, SimpleCacheClientBuilder},
+};
+
+use crate::error::CliError;
+
+pub async fn get_momento_client(auth_token: String) -> Result<SimpleCacheClient, CliError> {
+    SimpleCacheClientBuilder::new_with_explicit_agent_name(
+        auth_token,
+        NonZeroU64::new(100).unwrap(),
+        "cli",
+    )
+    .map_or_else(
+        |error| {
+            Err(CliError {
+                msg: error.to_string(),
+            })
+        },
+        |builder| Ok(builder.build()),
+    )
+}
+
+pub fn print_whatever_this_is_as_json<T>(value: &T)
+where
+    T: serde::Serialize,
+{
+    println!(
+        "{}",
+        serde_json::to_string_pretty(value).expect("Could not print whatever this is as json")
+    );
+}
+
+pub async fn interact_with_momento<U, FutureT>(
+    debug_note: &str,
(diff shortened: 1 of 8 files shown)


src/utils/client.rs
```diff
@@ -0,0 +1,49 @@
...
+
+pub fn print_whatever_this_is_as_json<T>(value: &T)
+where
+    T: serde::Serialize,
+{
+    println!(
+        "{}",
+        serde_json::to_string_pretty(value).expect("Could not print whatever this is as json")
+    );
+}
```
tylerburdsall, line 34: Minor nit: I like having a function that uses generics to pretty print results, could we name it to something more streamlined like `print_as_json`? `print_whatever_this_is` almost seems a little tongue-in-cheek for a public-vended CLI

tylerburdsall (commented):
Minor comment, everything else looks good. Thank you for helping reduce our copy-pasta code, this is super clean

src/utils/client.rs
kvcache: Sure, I'll remove this tiny island of joy from the sea of dispassion. I still need to look into windows too.

cprice404 (approved):
nice cleanup, ty

kvcache:
converted to draft because I need to mess with the github workflow to make windows work. I don't know why it can't find cmake.

kvcache:
Integration tests are broken for unrelated reasons.

cprice404 (approved):

bruuuuuuuce (approved):
:shipit:

tylerburdsall (approved):

PR #152, opened 2022-05-25
feat: add `login` command

kvcache:
Adds `momento login` command for doing SSO. The integration has
a live connection back up to the Momento endpoint that it running
the login workflow.

A standard oauth dance happens, with a browser opening up to an
SSO page. Some infos are printed from the server while you're
logging in, then the login credentials are printed to stdout.

We will be moving this session token to a momento dotfile, but
this change gets the SSO integration started. Not many endpoints
are supported, but over time we'll add more.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,8 @@ toml = "0.5.8"
 maplit = "1.0.2"
 configparser = "3.0.0"
 regex = "1"
-momento = "0.5.0"
+momento = "0.5.1"
+webbrowser = "0.7.1"
 
 [dev-dependencies]
 assert_cmd = "2.0.2"
--- a/src/commands/login.rs
+++ b/src/commands/login.rs
@@ -1,0 +1,23 @@
+use momento::momento::auth::LoginResult;
+
+pub async fn login() -> LoginResult {
+    momento::momento::auth::login(|action| {
+        match action {
+            momento::momento::auth::LoginAction::OpenBrowser(open) => {
+                // TODO: Need an early-out from the action sink.
+                // match webbrowser::open(&open.url) {
+                //     Ok(_) => {
+                //         log::debug!("opened browser to {}", open.url);
+                //     },
+                //     Err(e) => {
+                //         return LoginResult::NotLoggedIn(NotLoggedIn { error_message: e.to_string })
+                //     },
+                // }
+                webbrowser::open(&open.url).expect("Unable to open browser")
+            },
+            momento::momento::auth::LoginAction::ShowMessage(message) => {
+                println!("{}", message.text);
+            },
+        }
+    }).await
+}
--- a/src/commands/login.rs
+++ b/src/commands/login.rs
@@ -14,10 +14,11 @@ pub async fn login() -> LoginResult {
                 //     },
                 // }
                 webbrowser::open(&open.url).expect("Unable to open browser")
-            },
+            }
             momento::momento::auth::LoginAction::ShowMessage(message) => {
                 println!("{}", message.text);
-            },
+            }
         }
-    }).await
+    })
+    .await
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -42,8 +42,7 @@ enum Subcommand {
         operation: AccountCommand,
     },
     #[structopt(about = "Log in to manage your Momento account")]
-    Login {
-    }
+    Login {},
 }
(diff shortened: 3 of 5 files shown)


src/main.rs
```diff
@@ -41,6 +41,8 @@ enum Subcommand {
         #[structopt(subcommand)]
         operation: AccountCommand,
     },
+    #[structopt(about = "Log in to manage your Momento account")]
```
rlinehan, line 44: I wonder if we should maybe consider not adding this command to the help info until it is part of a usable workflow?

src/main.rs
```diff
@@ -251,6 +253,16 @@ async fn entrypoint() -> Result<(), CliError> {
                 commands::signingkey::signingkey_cli::list_signing_keys(creds.token).await?;
             }
         },
+        Subcommand::Login {} => match commands::login::login().await {
+            momento::momento::auth::LoginResult::LoggedIn(logged_in) => {
+                println!("{}", logged_in.session_token)
+            }
+            momento::momento::auth::LoginResult::NotLoggedIn(not_logged_in) => {
+                return Err(CliError {
+                    msg: format!("Failed to log in: {}", not_logged_in.error_message),
```
rlinehan, line 262: Nitpick... when I tried to do this pointed at prod, which still has `localhost` as the callback url and obviously couldn't be opened, I got:

```
ERROR: Failed to log in: Failed to log in: InternalServerError("error reading a body from connection: stream error received: not a result of an error")
```

The duplicate `Failed to log in` is a little silly?

src/commands/login.rs
```diff
@@ -0,0 +1,24 @@
+use momento::momento::auth::LoginResult;
+
+pub async fn login() -> LoginResult {
+    momento::momento::auth::login(|action| {
+        match action {
+            momento::momento::auth::LoginAction::OpenBrowser(open) => {
+                // TODO: Need an early-out from the action sink.
```
rlinehan, line 7: tbh I don't really understand what the phrase `need an early-out from the action sink` means (what's the 'action sink'), but I assume you're planning on following up on this?

rlinehan (commented):
looks good, pulled it down and it works! only real concern is whether we maybe want to hold off on including `login` in the help for now or make a note that it's still in progress, since you can't actually use the token it returns for anything yet.

src/commands/login.rs
kvcache: yes

src/main.rs
kvcache: k

src/main.rs
kvcache: structopt doesn't have hidden for this type. I'll add a note to the help text.

cprice404 (commented):
Will this be auto-released into the wild on PR merge or do we have a good mechanism for deciding when we are ready to release?

Cargo.toml
```diff
@@ -14,7 +14,8 @@ toml = "0.5.8"
 maplit = "1.0.2"
 configparser = "3.0.0"
 regex = "1"
-momento = "0.5.0"
+momento = "0.5.1"
+webbrowser = "0.7.1"
```
cprice404, line 18: this may have been mentioned in parking lot this morning, sorry if I wasn't paying close enough attention.

Do we know how a canary for this is going to work if it has a dep on browser interaction?

cprice404 (commented):
🙈 other than questions about canary->browser interaction, and how we control when this gets released.

Cargo.toml
kvcache: Nope! There might be a path with something like the Lynx browser but doing user-interactive identity provider selection is absolutely going to be interesting in the canary. `webbrowser` mentions Lynx support so seems like we won't be the very first to be possibly reaching for esoteric stuff.

Our canary might end up using a SAML IDP with different semantics than the social provider we're playing with now. Ruth is looking at how SAML assertions change/don't change the login flow.

kvcache:
> Will this be auto-released into the wild on PR merge or do we have a good mechanism for deciding when we are ready to release?

we are gung-ho zealous about just shipping `main` as though it is 100% always stable and ready to go. Main goes to homebrew in minutes.

Cargo.toml
cprice404: ACK

rlinehan (approved):

PR #154, opened 2022-06-02
feat: store session tokens in .momento_session

kvcache:
Also uses sessions if they're valid for at least 10 more seconds.
This breaks interactions with the service currently if you have
an active login session, because session tokens do not have the
momento endpoints baked into them.

### DIFF
--- a/src/commands/login.rs
+++ b/src/commands/login.rs
@@ -1,22 +1,24 @@
-use momento::momento::auth::LoginResult;
+use momento::{momento::auth::LoginResult, response::error::MomentoError};
 
 pub async fn login() -> LoginResult {
     momento::momento::auth::login(|action| {
         match action {
             momento::momento::auth::LoginAction::OpenBrowser(open) => {
-                // TODO: Need an early-out from the action sink.
-                // match webbrowser::open(&open.url) {
-                //     Ok(_) => {
-                //         log::debug!("opened browser to {}", open.url);
-                //     },
-                //     Err(e) => {
-                //         return LoginResult::NotLoggedIn(NotLoggedIn { error_message: e.to_string })
-                //     },
-                // }
-                webbrowser::open(&open.url).expect("Unable to open browser")
+                match webbrowser::open(&open.url) {
+                    Ok(_) => {
+                        log::debug!("opened browser to {}", open.url);
+                        None
+                    },
+                    Err(e) => {
+                        Some(
+                            Err(MomentoError::ClientSdkError(format!("Unable to open browser: {:?}", e)))
+                        )
+                    },
+                }
             }
             momento::momento::auth::LoginAction::ShowMessage(message) => {
-                println!("{}", message.text);
+                eprintln!("{}", message.text);
+                None
             }
         }
     })
--- a/src/commands/login.rs
+++ b/src/commands/login.rs
@@ -1,25 +1,22 @@
 use momento::{momento::auth::LoginResult, response::error::MomentoError};
 
 pub async fn login() -> LoginResult {
-    momento::momento::auth::login(|action| {
-        match action {
-            momento::momento::auth::LoginAction::OpenBrowser(open) => {
-                match webbrowser::open(&open.url) {
-                    Ok(_) => {
-                        log::debug!("opened browser to {}", open.url);
-                        None
-                    },
-                    Err(e) => {
-                        Some(
-                            Err(MomentoError::ClientSdkError(format!("Unable to open browser: {:?}", e)))
-                        )
-                    },
+    momento::momento::auth::login(|action| match action {
+        momento::momento::auth::LoginAction::OpenBrowser(open) => {
+            match webbrowser::open(&open.url) {
+                Ok(_) => {
+                    log::debug!("opened browser to {}", open.url);
+                    None
                 }
+                Err(e) => Some(Err(MomentoError::ClientSdkError(format!(
+                    "Unable to open browser: {:?}",
+                    e
+                )))),
             }
-            momento::momento::auth::LoginAction::ShowMessage(message) => {
-                eprintln!("{}", message.text);
-                None
-            }
+        }
+        momento::momento::auth::LoginAction::ShowMessage(message) => {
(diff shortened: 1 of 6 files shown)


src/utils/user.rs
```diff
@@ -1,44 +1,78 @@
...
-    let creds = match get_creds_for_profile(profile).await {
-        Ok(c) => c,
-        Err(e) => return Err(e),
-    };
-    let config = match get_config_for_profile(profile).await {
-        Ok(c) => c,
-        Err(e) => return Err(e),
-    };
+    let creds = get_creds_for_profile(profile).await?;
+    println!("CREDS {:?}", creds);
```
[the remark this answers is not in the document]
kvcache: woops

tylerburdsall (approved):
Looks good, might need to be formatted

src/commands/login.rs
```diff
@@ -1,22 +1,24 @@
...
-                // }
-                webbrowser::open(&open.url).expect("Unable to open browser")
+                match webbrowser::open(&open.url) {
+                    Ok(_) => {
+                        log::debug!("opened browser to {}", open.url);
+                        None
+                    },
+                    Err(e) => {
+                        Some(
+                            Err(MomentoError::ClientSdkError(format!("Unable to open browser: {:?}", e)))
```
rlinehan, line 14: I wonder if it would be good in this case to print out what url the browser would have opened to so users can manually open it?

src/utils/user.rs
```diff
@@ -1,44 +1,77 @@
...
+fn set_session_token(credentials: &mut Ini, session_token: Option<String>, valid_for_seconds: u32) {
+    let expiry_time = Utc::now() + Duration::seconds(valid_for_seconds.into());
+    credentials.set(".momento_session", "token", session_token);
+    credentials.set(".momento_session", "valid_until", Some(expiry_time.timestamp().to_string()));
+}
+
+pub async fn clobber_session_token(session_token: Option<String>, valid_for_seconds: u32) -> Result<(), CliError> {
+    let mut credentials_file = read_credentials().await?;
+    set_session_token(&mut credentials_file, session_token, valid_for_seconds);
+    ini_write_to_file(credentials_file, &get_credentials_file_path()).await?;
```
rlinehan, line 43: I wonder if it would be worth a debug statement here about where it's being written to? (mostly cuz I ended up adding this myself while testing this out...)

src/utils/user.rs
```diff
@@ -1,44 +1,77 @@
...
-                .to_string(),
-        }),
-    };
+    let credentials_file = read_credentials().await?;
 
-    let creds_result = match credentials.get(profile, "token") {
-        Some(c) => c,
-        None => return Err(CliError{
+    get_session_token(&credentials_file).or_else(|| {
+        credentials_file.get(profile, "token")
```
rlinehan, line 58: Ok so just to make sure I'm understanding correctly, any session tokens will take precedence over api tokens? My only thought is (in the future) we might want at least some debug logging about what token/profile is actually in use so that users have some visibility into that, otherwise I can see some confusing scenarios (when we get more into authorization) where a user thinks they have an active session that has more permissions than whatever api token is in place, but actually that session has expired.

src/utils/user.rs
```diff
@@ -1,44 +1,77 @@
...
+            .map(|expiry_timestamp| { Utc.timestamp(expiry_timestamp, 0) });
+        if let Some(expiry_timestamp) = expiry {
+            if Utc::now() + Duration::seconds(10) < expiry_timestamp {
+                let expiring = expiry_timestamp - Utc::now();
+                log::debug!("Found user session expiring in {}m", expiring.num_minutes());
+                return session_token;
+            }
+            log::debug!("Token already expired at: {}", expiry_timestamp);
+        } else {
+            log::debug!(".momento_session profile is missing the expiry time. Skipping this session...");
```
rlinehan, line 27: I think I had originally thought that users would have the ability to `momento login` to separate profiles, rather than a session being a profile under the hood. I realize now that there isn't really per-profile information at the `login` level - that is, you log in to all of `momento`, then do different things with different profiles, e.g. have projects and caches.

I think that all works at least for now... my only concern is what if we add authorization - would you ever have a case where you have different `login`s that are associated with different levels/areas of access, and so you want to be able to have multiple currently active session tokens? I'm not sure we will, especially if we give API tokens different levels of access, but just something I was thinking about. Also, that makes me wonder if we'll want a `momento logout` too.

rlinehan (dismissed):
Looks good! Left a couple comments about how the `.momento_session` works (and how session tokens and api tokens might interact), but none of that is blocking and also might be easier over chat or something.

src/commands/login.rs
kvcache: Hmm, maybe. 🤔 I'll see if I can come up with something better here - I've got a few ideas to try.

src/utils/user.rs
kvcache: I didn't need any more than this; if you have a line you want inserted, feel free to propose a statement you find useful and a location - or open a pr!

src/utils/user.rs
kvcache: Yeah. I haven't put together that precedence scheme yet; currently trying to get the system doing the basic thing and this profile stuff has some rough edges I don't want to block on straightening out.

src/utils/user.rs
kvcache: You have 1 identity for `momento` and have access to quite a few accounts and roles; afaict that is not a limiting factor to your developer authorization experience right? I think it's unclear that we need to model identity rules _in addition_ to rules about what identities can do (which we need regardless)

I was thinking about `momento logout` too; should be simple, we can remove the token from the session section. Later pr, we can add it to the board

rlinehan: > You have 1 identity for `momento` and have access to quite a few accounts and roles

Yep, I think that makes sense here. I was thinking some about how AWS does it where you can tell different profiles how to do sso, but if we ever did need that I think we could add it later.

PR #155, opened 2022-06-03
feat: qr code login option

kvcache:
`momento login qr`

Prints a unicode QR image for you to log in via your phone.
`momento login` defaults to `momento login browser`.

### DIFF
--- a/src/commands/login.rs
+++ b/src/commands/login.rs
@@ -1,7 +1,24 @@
+use momento::momento::auth::{EarlyOutActionResult, LoginAction};
 use momento::{momento::auth::LoginResult, response::error::MomentoError};
+use qrcode::render::unicode;
+use qrcode::QrCode;
 
-pub async fn login() -> LoginResult {
-    momento::momento::auth::login(|action| match action {
+#[derive(clap::ArgEnum, Clone, Debug)]
+pub enum LoginMode {
+    Browser,
+    Qr,
+}
+
+pub async fn login(login_mode: LoginMode) -> LoginResult {
+    momento::momento::auth::login(match login_mode {
+        LoginMode::Browser => login_with_browser,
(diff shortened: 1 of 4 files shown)


src/commands/login.rs
```diff
@@ -18,6 +35,25 @@ pub async fn login() -> LoginResult {
...
         }
-    })
-    .await
+    }
+}
+
+fn login_with_qr_code(action: LoginAction) -> EarlyOutActionResult {
+    match action {
+        momento::momento::auth::LoginAction::OpenBrowser(open) => {
+            eprintln!("Navigate here to log in: {}", open.url);
```
tylerburdsall, line 44: Shouldn't this be `println!`? I thought `eprintln` was only for `stderr`. Or is this some workaround to get the QR code to actually show up?

src/commands/login.rs
```diff
@@ -18,6 +35,25 @@ pub async fn login() -> LoginResult {
...
+            let code = QrCode::new(open.url).unwrap();
+            let image = code
+                .render::<unicode::Dense1x2>()
+                .dark_color(unicode::Dense1x2::Dark)
+                .light_color(unicode::Dense1x2::Light)
+                .build();
+            eprintln!("{}", image);
```
tylerburdsall, line 51: Non-blocking, curious to know how this would appear on light-theme terminal background

tylerburdsall (commented):
Just a couple comments, everything is looking good though!

bruuuuuuuce (approved):
:shipit:

tylerburdsall (approved):

src/commands/login.rs
kvcache: image

tylerburdsall: Nice! Cool to see that it handles that

PR #159, opened 2022-07-07
feat: changes name of delete command

### LINKED ISSUE #145: Add Delete functionality
Our backend server now supports delete, this functionality needs to be added into the CLI.

Need to update the rust sdk first https://github.com/momentohq/client-sdk-rust/issues/49

WillRInternship:
Closes #145 

```console
$./momento cache
momento-cache 
Interact with caches

USAGE:
    momento cache [OPTIONS] <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
        --verbose    

SUBCOMMANDS:
    create-cache    Create a cache
    delete-cache    Delete the cache
    get             Get an item from the cache
    help            Print this message or the help of the given subcommand(s)
    list-cache      List all caches
    set             Store a given item in the cache
```


src/main.rs
```diff
@@ -110,7 +110,7 @@ enum CloudSignupCommand {
 #[derive(Debug, StructOpt)]
 enum CacheCommand {
     #[structopt(about = "Create a cache")]
-    Create {
+    CreateCache {
```
kvcache, line 113: so this would make people need to type
```
momento cache createcache --name cachename
```
right? I'm not convinced that writing cache again right after writing cache is removing any lack of clarity that already exists. Can you explain the motivation behind this change?

eaddingtonwhite: @kvcache he was going directly off this ticket per @danielamiao comments. Im updating pr w/ ticket link as well now.
https://github.com/momentohq/momento-cli/issues/145

I know we had talked as well to collapse commands down to root of CLI but would rather that be follow up pr then after this initial refactor for ticket @WillRInternship working on now.

eaddingtonwhite: also this boils down to `momento cache create-cache --name cachename` with how clap parses commands currently just FYI it adds in the `-`

eaddingtonwhite: ok we clarified w/ @danielamiao we want to move down to root. @WillRInternship  check out comment here.
https://github.com/momentohq/momento-cli/issues/145#issuecomment-1178053458

Id still from implementation wise get delete added first make sure works as you would expect. Then can make a follow up commit to get rid of the `cache` subcommand and move all the commands down to root of CLI like @danielamiao saying in ticket.

poppoerika:
Closing this pr for now and will come back to work on the updates.
cc @cprice404

PR #161, opened 2022-07-21
feat: Add endpoint flag option

gautamomento:
Adding a `--endpoint` option that would allow to specify what momento endpoint to use for cli operations to connect to.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -818,9 +818,9 @@ dependencies = [
 
 [[package]]
 name = "momento"
-version = "0.5.4"
+version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


gautamomento:
Punting Config based override to https://github.com/momentohq/momento-cli/issues/162

rlinehan (commented):
This looks good to me, although I still haven't done a ton of Rust.

Only question is are there any tests we should add to ensure that using an `endpoint` doesn't cause an error?

gautamomento:
> This looks good to me, although I still haven't done a ton of Rust so might be good to get Kenny's input too.
> 
> Only question is are there any tests we should add to ensure that using an `endpoint` doesn't cause an error?

Yup, they are right here: https://github.com/momentohq/momento-cli/blob/main/tests/momento_default_profile.rs and they seem to be passing.

rlinehan (approved):

PR #169, opened 2022-10-11
feat: bundle login as separate feature

### LINKED ISSUE #167: CLI - hide "under construction" config option

malandis:
The login option is still actively under construction. In order to separate this from other stable features, we have classified the login functionality as a conditionally compiled feature. By default login will not be included, but may be adding `--features login` to `cargo build` or `cargo run`.

Closes #167

### DIFF
--- a/src/utils/user.rs
+++ b/src/utils/user.rs
@@ -7,6 +7,7 @@ use crate::{
     utils::file::{get_config_file_path, get_credentials_file_path, read_file},
 };
 
+#[cfg(feature = "login")]
 use super::file::ini_write_to_file;
 
 fn get_session_token(credentials: &Ini) -> Option<String> {
@@ -33,6 +34,7 @@ fn get_session_token(credentials: &Ini) -> Option<String> {
     None
 }
 
+#[cfg(feature = "login")]
 fn set_session_token(credentials: &mut Ini, session_token: Option<String>, valid_for_seconds: u32) {
     let expiry_time = Utc::now() + Duration::seconds(valid_for_seconds.into());
     credentials.set(".momento_session", "token", session_token);
@@ -43,6 +45,7 @@ fn set_session_token(credentials: &mut Ini, session_token: Option<String>, valid
     );
 }
 
+#[cfg(feature = "login")]
 pub async fn clobber_session_token(
     session_token: Option<String>,
     valid_for_seconds: u32,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,10 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        feature: [default, login]
+
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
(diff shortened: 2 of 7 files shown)


cprice404:
@malandis do the list of "features" show up in the help somewhere?  If so, I would prefer to just comment out the option.  If not, then this is probably fine.

@rlinehan would you mind reviewing this?  My suggestion was to just comment out the subcommand for now, I figure when you all start picking this work up again you can just use a local build of `mm` until the command is ready for prime time.  I don't really want the "under construction" thing to show up in the help menu after our GA announcement.  Just wanna make sure this is okay with you.

src/utils/user.rs
```diff
@@ -7,6 +7,7 @@ use crate::{
     utils::file::{get_config_file_path, get_credentials_file_path, read_file},
 };
 
+#[cfg(feature = "login")]
```
cprice404, line 10: I don't totally understand what this macro does, but I think we do `ini` file writes for the `configure` subcommand in addition to the login one... this change doesn't preclude `configure` from working properly does it?

malandis:
> @malandis do the list of "features" show up in the help somewhere? If so, I would prefer to just comment out the option. If not, then this is probably fine.

No, it doesn't show up. This is more like a feature flag. You have to specify you want it included at build time, eg

`cargo build --features login`

Otherwise it is not compiled in

src/utils/user.rs
malandis: > I don't totally understand what this macro does, but I think we do ini file writes for the configure subcommand in addition to the login one... this change doesn't preclude configure from working properly does it?

It's conditionally compiling the `use` statement for just this one module, since `file::ini_write_to_file` is only used in this module by `clobber_session_token`. We use `file::ini_write_to_file` in other modules though.

rlinehan (approved):
I haven't done a ton with rust, so can't comment a ton on correctness (might be good to get @kvcache or @gautamomento to look at this, as both of them have done more work on the login command), but I really like this solution of using a feature flag rather than commenting things out - as @malandis noted in slack, if we comment out some parts then we'll end up with compiler errors about unused functions and will have to comment out a lot more. This feels like much less work overall to be able to continue iterating on the login command easily without publishing it.

gautamomento (commented):
This looks good to me.

I like that this will not be bundled when the cli gets installed on customer's machines. I like that it grants a good amount of  developer flexibility rather than having to maintain feature branches. 

However, I couldn't figure out if the features are included in build steps when we open PRs or merge PRs? I would definitely want that the features can be compiled.

malandis:
> However, I couldn't figure out if the features are included in build steps when we open PRs or merge PRs? I would definitely want that the features can be compiled.

Thank you for considering this. In b0fc3893cbf438adf232f0890a49e8abe83625e9 I added a matrix to the CI workflow so that the build and tests will exercise over both the production features and also the login features. That way when a developer works on the login code they will also have automated build and tests through GH. Once we graduate `login` to stable we can remove the feature from both the code and the CI workflow.

gautamomento (approved):
LGTM. Thank you for doing this.

rlinehan (approved):
Love the test matrix! :shipit:

cprice404 (approved):

PR #170, opened 2022-10-12
chore: refactor profile option to be allowed in any position

### LINKED ISSUE #99: `--profile` should be a global flag (via clap)
flags that are configured as "global" should be able to appear in any order in the CLI input.  For example, I expect to be able to write this:

```
momento cache --profile my-favorite-profile get --key foo
```

but I get an error saying that I have to put `--profile` after the subcommand `get`.  I *think* that we have some arguments like `--verbose` in our internal rust CLI tools that are set to "global" and thus it will allow you to put them in any order in the CLI command.  This would be really useful for `--profile` in particular because I'm likely to run multiple different `get` / `set` commands and I want to be able to use my zsh history without having to backspace over the entire `--profile` string to get to the part of the command that I'm trying to change.

malandis:
This refactors the CLI arguments so that `profile` may appear in any position. Previously `profile` was specific to the leaf subcommands. Under this setup, a user had to write:

`momento cache get --key key --profile my-profile`

but could not write either of:

`momento cache --profile my-profile get --key key`
`momento --profile my-profile cache get --key key`

Because users will expect all these variations, we change `profile` to be a global option.

Closes #99

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -396,6 +396,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "279fb028e20b3c4c320317955b77c5e0c9701f05a1d309905d6fc702cdc5053e"
 
 [[package]]
+name = "float-cmp"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


bruuuuuuuce (approved):
:shipit:

cprice404 (approved):
perfect

PR #171, opened 2022-10-13
fix: display available regions for aws/gcp for better error handling

### LINKED ISSUE #168: CLI - better error handling / help messages for regions
Provide better messages when:

1. users didn't pass a value for `--region` for `momento account signup` (see current message below)
```
momento account signup aws --email 123@email.com --region
error: The argument '--region <REGION>' requires a value but none was supplied
```

2. users didn't pass `--region` flag and its value for `momento account signup` (see current message below)
```
momento account signup gcp --email 123@email.com                
error: The following required arguments were not provided:
    --region <REGION>
```

poppoerika:
Closes #168 

### When user didn't pass `--region` flag; available regions are displayed 

Before
```
momento account signup gcp --email eem@momentohq.com
error: The following required arguments were not provided:
    --region <REGION>
```
After
```
target/debug/momento account signup gcp --email eem@momentohq.com
error: The following required arguments were not provided:
    --region <us-east1 or ap-northeast1>
```

### When user passed an invalid value for `--region`; possible values are displayed. Also, does not alert the cli-signup channel on Slack

Before
```
momento account signup aws --email eem@momentohq.com --region fgfs
[2022-10-13T04:37:39Z INFO  momento::commands::account] Signing up for Momento...
ERROR: Failed to create Momento token: Region 'fgfs' is not a valid region for AWS. The available regions are: us-west-2, us-east-1, ap-northeast-1
```
After
```
target/debug/momento account signup gcp
--email eem@momentohq.com --region dcasdfafaf
error: "dcasdfafaf" isn't a valid value for '--region <us-east1 or ap-northeast1>'
	[possible values: ap-northeast1, us-east1]
```

### When user didn't pass any value to `--region` flag; availabe regions are displayed in the error message

Before
```
momento account signup aws --email eem@momentohq.com --region
error: The argument '--region <REGION>' requires a value but none was supplied
```
After
```
target/debug/momento account signup gcp --email eem@momentohq.com --region
error: The argument '--region <us-east1 or ap-northeast1>' requires a value but none was supplied
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -101,14 +101,14 @@ enum CloudSignupCommand {
     Gcp {
         #[structopt(long, short)]
         email: String,
-        #[structopt(long, short, help = "e.g. us-east1, ap-northeast1")]
+        #[structopt(long, short, possible_values = ["us-east1", "ap-northeast1"], value_name = "us-east1 or ap-northeast1")]
         region: String,
     },
     #[structopt(about = "Signup for Momento on AWS")]
     Aws {
         #[structopt(long, short)]
         email: String,
-        #[structopt(long, short, help = "e.g. us-west-2, us-east-1, ap-northeast-1")]
+        #[structopt(long, short, possible_values = ["us-west-2", "us-east-1", "ap-northeast-1"], value_name = "us-west-2, us-east-1, or ap-northeast-1")]
         region: String,
     },
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -101,14 +101,14 @@ enum CloudSignupCommand {
     Gcp {
         #[structopt(long, short)]
         email: String,
-        #[structopt(long, short, possible_values = ["us-east1", "ap-northeast1"], value_name = "us-east1 or ap-northeast1")]
+        #[structopt(long, short, value_name = "us-east1 or ap-northeast1")]
         region: String,
     },
     #[structopt(about = "Signup for Momento on AWS")]
     Aws {
         #[structopt(long, short)]
         email: String,
-        #[structopt(long, short, possible_values = ["us-west-2", "us-east-1", "ap-northeast-1"], value_name = "us-west-2, us-east-1, or ap-northeast-1")]
+        #[structopt(long, short, value_name = "us-west-2, us-east-1, or ap-northeast-1")]
         region: String,
     },
 }


src/main.rs
```diff
@@ -101,14 +101,14 @@ enum CloudSignupCommand {
...
-        #[structopt(long, short, help = "e.g. us-east1, ap-northeast1")]
+        #[structopt(long, short, possible_values = ["us-east1", "ap-northeast1"], value_name = "us-east1 or ap-northeast1")]
         region: String,
     },
     #[structopt(about = "Signup for Momento on AWS")]
     Aws {
         #[structopt(long, short)]
         email: String,
-        #[structopt(long, short, help = "e.g. us-west-2, us-east-1, ap-northeast-1")]
+        #[structopt(long, short, possible_values = ["us-west-2", "us-east-1", "ap-northeast-1"], value_name = "us-west-2, us-east-1, or ap-northeast-1")]
```
gautamomento, line 111: Thank you for taking a stab at this. If I understand this correctly, the CLI would be enforcing these strings and will prevent any other region strings from being passed. 

While this will work correctly, it would mean that each time we launch a new region customers will not be able to signup to the new regions till we publish a new version of CLI with updated string in this section and they must update their CLI to use the new region. Just wanted to make sure that we want to switch over to this behavior.

This is better experience for our customers to see in terms of what regions are available upfront without a roundtrip to our service. But just wanted to call out that this is coming in at cost of some flexibility.

gautamomento, line 111: Thank you for taking a stab at this. If I understand this correctly, the CLI would be enforcing these strings and will prevent any other region strings from being passed. 

While this will work correctly, it would mean that each time we launch a new region customers will not be able to signup to the new regions till we publish a new version of CLI with updated string in this section and they must update their CLI to use the new region. Just wanted to make sure that we want to switch over to this behavior.

This is better experience for our customers to see in terms of what regions are available upfront without a roundtrip to our service. But just wanted to call out that this is coming in at cost of some flexibility.

gautamomento (commented):
Thank you for working on making the customer experience better. I have a left a comment for you to consider.

src/main.rs
cprice404: yes, @gautamomento is correct.  Sorry @poppoerika, I was AFK yesterday so didn't get a chance to talk through this with you.

I think we can't restrict to `possible_values` in case new regions become available on the server.

What does the `value_name` attribute do?  Does that just affect the help/error messages?  We might be able to keep that one!

You have 3 examples in your (excellent) PR description.  If possible we want to keep the improvements to the help messages that you showed in your first and third example.  But for the second one we want to allow the request to go up to the server and then print the response from the server, as it was doing prior to your PR.

Let me know if any of that doesn't make sense!

poppoerika: @gautamomento, @cprice404 
Thank you for the feedback!

> I think we can't restrict to possible_values in case new regions become available on the server.

Understood. My original understanding was to not alerting the cli-signup channel with the wrong region name during signup, but it makes sense that we don't want to release a new version CLI every time we add a new region and keep both in sync.

> What does the value_name attribute do? Does that just affect the help/error messages? We might be able to keep that one!

Yup. `value_name` will be displayed following the flag name.
So before it was `--region <REGION>` (not really helpful. 
But now it's like `--region <us-west-2, us-east-1, or ap-northeast-1>`
This should not affect a request going up to the server with the wrong region name so we want to keep it, I believe.

poppoerika: But we need to update `value_name` as a new region is added to provide accurate information.

cprice404: Yep!  Understood that the CLI will need to be updated if we want the list to be authoritative on the client side.  But we should still have a goal that if a user has an older version of the CLI and we tell them that a new region is available, they will still be able to sign up.

bruuuuuuuce (approved):
:shipit:

PR #173, opened 2022-10-13
chore: try and solve release race condition

bruuuuuuuce:
hopefully fixes https://github.com/momentohq/oncall-tracker/issues/136 🤞

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -36,13 +36,14 @@ jobs:
         with:
           python-version: "3.x"
 
-      - name: Set release
+      - name: Dry run release
         id: semrel
         uses: go-semantic-release/action@v1
         with:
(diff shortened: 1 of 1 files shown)


mjindalmotiv8 (approved):

.github/workflows/cd.yml
```diff
@@ -96,6 +97,14 @@ jobs:
       - name: Output release
```
cprice404, line 97: so the `semrel` step will still produce `outputs` even when run in `dry` mode?

bruuuuuuuce: yep I verified that here https://github.com/momentohq/momento-local/actions/runs/3245284851/jobs/5322611286

cprice404 (approved):
race

PR #175, opened 2022-10-14
chore: only use logging when verbose

### LINKED ISSUE #172: CLI - audit log messages that can show up on STDOUT, make sure the logging implementation isn't too visible

malandis:
This turns off logging when not verbose. Verbose logging is set to `debug` level. As discussed in #172 we find the log data _too_ verbose compared to the other status messages or normal cache response data.

We replace other print statements with one of two new macros.

We funnel all status or error print statements to `stderr` with `console_info!`. Similarly we define a macro `console_data!` for cache response data, which we print to `stdout`.

Closes #172

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -5,7 +5,8 @@ use clap::StructOpt;
 use commands::login::LoginMode;
 use env_logger::Env;
 use error::CliError;
-use log::{debug, error};
+use log::{debug, error, LevelFilter};
+use utils::console::console_info;
 use utils::user::get_creds_and_config;
 
 #[cfg(feature = "login")]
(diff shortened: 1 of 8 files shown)


cprice404 (approved):
This looks wonderful, thanks!

src/main.rs
```diff
@@ -185,7 +186,12 @@ enum CacheCommand {
 async fn entrypoint() -> Result<(), CliError> {
     let args = Momento::parse();
 
-    let log_level = if args.verbose { "debug" } else { "info" };
+    let log_level = if args.verbose {
+        LevelFilter::Debug
+    } else {
+        LevelFilter::Off
```
kvcache, line 192: this broke `panic!()` output.

Issue #177, opened 2022-10-17
Fix release process


cprice404:

cprice404:
Work for this ticket is probably relevant for #181 as well.

poppoerika:
<img width="965" alt="image" src="https://user-images.githubusercontent.com/22686260/199323428-c2214473-c115-4e3d-b8e4-67e7063b9705.png">

PR #180, opened 2022-10-21
fix: add release step to pass release version

poppoerika:
This is to fix the test failure on the `on-push-to-main` workflow. 
A release version argument was missing from the `update-cargo` job, so added dry run release step to generate a release version and pass it to the `set_cargo_version` script.
Also added a step to set up python to run the `set_cargo_version` script successfully. 

Ticket #177

### DIFF
--- a/.github/workflows/on-push-to-main.yml
+++ b/.github/workflows/on-push-to-main.yml
@@ -87,15 +87,23 @@ jobs:
     runs-on: ubuntu-latest
     needs: [rustfmt, build, build-windows]
     steps:
+      - name: Dry run release
+        id: semrel
+        uses: go-semantic-release/action@v1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          allow-initial-development-versions: true
+          force-bump-patch-version: true
+          dry: true
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/on-push-to-main.yml
```diff
@@ -87,15 +87,27 @@ jobs:
...
       - name: Update Cargo Version
         run: |
           chmod +x set_cargo_version.sh
-          ./set_cargo_version.sh ${{ needs.release.outputs.version }}
+          ./set_cargo_version.sh ${{ steps.semrel.outputs.version }}
           git config user.email "momentobot@users.noreply.github.com"
           git config user.name "momentobot"
           git add Cargo.toml
-          git commit -m "chore: bump cargo version v${{ needs.release.outputs.version }}"
+          git commit -m "chore: bump cargo version v${{ steps.semrel.outputs.version }}"
```
cprice404, line 110: @poppoerika actually now that I look at this code, I think we don't want to actually bump the cargo version at all on push to main.  we only want to do that on push to `release`.  So can you try just removing this action (line 102) and also the "push" action (line 112)?

cprice404 (approved):

PR #187, opened 2022-11-01
fix: release before building tar.gz file

poppoerika:
Since we're doing dry-run before building a tar.gz file, release id was not retrieved properly, which resulted in `null`.
Therefore, the upload asset API failed. 

I'm not fully sure if this fixes the homebrew repo issue, but the last success we had matched what I'm pushing in this pr.

### DIFF
--- a/.github/workflows/on-push-to-release.yml
+++ b/.github/workflows/on-push-to-release.yml
@@ -35,14 +35,13 @@ jobs:
         with:
           python-version: "3.x"
 
-      - name: Dry run release
+      - name: Set release
         id: semrel
         uses: go-semantic-release/action@v1
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/on-push-to-release.yml
```diff
@@ -35,14 +35,13 @@ jobs:
         with:
           python-version: "3.x"
 
-      - name: Dry run release
+      - name: Set release
```
cprice404, line 38: okay, you're right, we definitely need this.  ultimately i think we might want to do:

* dry run to get the next version number
* update the cargo version number and commit that back to the branch
* then to the full semver to actually tag it, including the updated cargo version number

But we can merge this one for now to see if we can just get a release out, and then work on that ^^ as a follow-up.

cprice404 (approved):

Issue #181, opened 2022-10-31
on linux, momento cli reports the wrong version number


cprice404:
Version 0.22.4 is reporting itself as 0.11.9, need to make sure the cargo version is getting updated properly prior to releases.

PR #188, opened 2022-12-02
fix: add one more param to `needs` in publishing assets jobs on release

poppoerika:
Ran the latest momento-cli on an Amazon Linux EC2 instance, and the version was not the oldest `0.11.9` as described in [this ticket](https://github.com/momentohq/momento-cli/issues/181), but it was off by one minor version (supposed to be `0.22.6` but it was `0.22.5`).
I believe this is happening because the cli is built and released before the cargo version is updated and pushed to `release` branch so when the cli is build and released on the publishing assets jobs, it's still referencing the previous version listed in `cargo.toml`.
I added `update-cargo` as one of the needed jobs for both publishing jobs, so hoping that would fix the above issue.

Once this is merged and run the manual release and verify its version on the Linux machine, I will close the ticket: https://github.com/momentohq/momento-cli/issues/181

### DIFF
--- a/tests/configure_profiles_test.rs
+++ b/tests/configure_profiles_test.rs
@@ -6,7 +6,7 @@ mod tests {
     async fn configure_momento_default_profile() {
         let test_auth_token = std::env::var("TEST_AUTH_TOKEN_DEFAULT").unwrap();
         let mut cmd = Command::cargo_bin("momento").unwrap();
-        cmd.args(&["configure"])
+        cmd.args(["configure"])
             .write_stdin(test_auth_token)
             .assert()
             .success();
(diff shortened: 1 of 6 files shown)


tests/configure_profiles_test.rs
```diff
@@ -6,30 +6,30 @@ mod tests {
     async fn configure_momento_default_profile() {
         let test_auth_token = std::env::var("TEST_AUTH_TOKEN_DEFAULT").unwrap();
         let mut cmd = Command::cargo_bin("momento").unwrap();
-        cmd.args(&["configure"])
+        cmd.args(["configure"])
```
schwern, line 9: Are these changes related? Or the result of an automatic process to remove an unnecessary reference?

poppoerika: Those changes were suggested from clippy. 
So I fixed them by running `cargo clippy --fix`.

schwern: I think it broke a test.

tests/momento_default_profile.rs
```diff
@@ -48,15 +48,15 @@ mod tests {
         let mut test_cache_default = std::env::var("TEST_CACHE_DEFAULT").unwrap();
         test_cache_default.push('\n');
         let mut cmd = Command::cargo_bin("momento").unwrap();
-        cmd.args(&["cache", "list"])
+        cmd.args(["cache", "list"])
             .assert()
             .stdout(test_cache_default);
```
schwern, line 53: I think this is the broken test.

poppoerika: Yup, I need to do some stuff first and will push a fix.

cprice404 (approved):

PR #196, opened 2022-12-08
feat: windows installer

malandis:
This builds a Windows installer for the Momento CLI. We use the Wix
packaging tooling. The configured installer allows the user to choose
an install location; the installer also updates the path environment
variable so the CLI is accessible at the command prompt, includes a
license, and an icon for the uninstall page.

We update the GitHub actions to build the installer and distribute it
for release.

A remaining step here is to sign the installer.


.github/workflows/on-push-to-release.yml
```diff
@@ -245,22 +245,28 @@ jobs:
...
         id: create_release
         run: |
-          $env:Version = "${{ needs.release.outputs.version }}"
-          $env:BinaryFile64 = "momento-cli-" + $env:VERSION + ".windows_x86_64.zip"
           rustup target add x86_64-pc-windows-gnu
           cargo build --release --target x86_64-pc-windows-gnu
-          $env:CurrentDir = Get-Location
-          $env:LiteralPath = $env:CurrentDir + "\target\x86_64-pc-windows-gnu\release\momento.exe"
-          $env:DestinationPath = $env:CurrentDir + "\" + $env:BinaryFile64
-          Compress-Archive -LiteralPath $env:LiteralPath -DestinationPath $env:DestinationPath
```
cprice404, line 258: rather than replacing the zip with the installer can we just include both of them in the release for now?  @pgautier404 can pick up this change though

cprice404 (approved):

PR #198, opened 2022-12-09
fix:`on-push-to-release` workflow

poppoerika:
I found out that the container used for the `publish-linux-asset` job, `quay.io/pypa/manylinux2014_x86_64`, does not work well with `actions/setup-python@v2`.
Therefore, I removed that action and installed python3 and pip3 with yum.
Tested locally with the exact same image, `quay.io/pypa/manylinux2014_x86_64`, and I was able to run `set_cargo_version.sh` successfully.
I also removed the container from the `publish-rpm-package` job to remove any complexity since we don't need to use a container to run this job.

This PR should make the `on-push-to-release` workflow run successfully and update the momento-cli version; therefore,
1. linux asset will be successfully released
2. correct momento-cli version will be displayed on linux
3. rpm package (x86_64) will be released 
🤞 

@cprice404 @virratanasangpunth 
cc/ @rlinehan @ksmotiv8

### DIFF
--- a/.github/workflows/on-push-to-release.yml
+++ b/.github/workflows/on-push-to-release.yml
@@ -136,8 +136,6 @@ jobs:
     strategy:
       matrix:
         architecture: [ x86_64 ]
-    container:
-      image: quay.io/pypa/manylinux2014_${{ matrix.architecture }}
     steps:
       - uses: actions/checkout@v2
       - name: Setup Python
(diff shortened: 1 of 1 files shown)


.github/workflows/on-push-to-release.yml
```diff
@@ -136,8 +136,6 @@ jobs:
     strategy:
       matrix:
         architecture: [ x86_64 ]
-    container:
-      image: quay.io/pypa/manylinux2014_${{ matrix.architecture }}
```
virratanasangpunth, line 140: code change looks fine. but let's not merge this tonight since the issue has already been mitigated. I don't wanna risk impacting the windows binary as well, which appears to be included in this workflow.

ksmotiv8:
Good call

virratanasangpunth (approved):
LGTM

I'm not familiar with this code so maybe let's wait for @cprice404 or @bruuuuuuuce to give you another +1.

cprice404 (approved):

PR #200, opened 2022-12-09
fix: no yum on ubuntu, try to pass GH auth header for Windows

cprice404:
This commit attempts to fix two current failures in the release
process:

1. The rpm release is failing because we try to do a `yum install jq`
   on ubuntu, which hopefully shouldn't be necessary, i am expecting
   the ubuntu image already has jq.
2. The windows install fails due to missing github auth token when
   making an API call t get the release url.


bruuuuuuuce (approved):
:shipit:

PR #202, opened 2022-12-10
fix: working on rpm, deb, and arm builds

cprice404:
This commit changes several things as we work toward finishing
out linux support (rpms, debs, and ARM builds):

* Attempt to do the Cargo.toml file update only once, at the
  beginning.
* Switches off of jiro4989/build-rpm-action in favor of using nfpm;
  this is because I tested the rpms generated by jiro4989 and they
  did not work on most of the RHEL platforms (such as AL2); I have
  tested rpms and debs built by nfpm on ubuntu, debian, centos,
  rhel, and AL2 and they work on all of them.
* Adds ARM into the arch matrix.

This is going to fail, so I left a lot of the previously working
code in place and commented out in case we need to refer back to
it as we are working out the kinks.  I will clean up all of the
commented code once things are working.

### DIFF
--- a/.github/workflows/on-push-to-release.yml
+++ b/.github/workflows/on-push-to-release.yml
@@ -24,7 +24,7 @@ jobs:
   release:
     runs-on: ubuntu-latest
     outputs:
-      version: ${{ steps.release.outputs.release }}
+      version: ${{ steps.semrel.outputs.release }}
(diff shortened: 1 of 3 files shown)


bruuuuuuuce (approved):
:shipit:

PR #203, opened 2022-12-13
chore: revert 39f71a3

malandis:
Revert change to README regarding Linux install location.


cprice404:
let's not merge this yet.  I may break things again before the dust settles.

PR #207, opened 2022-12-14
chore: arch string is different for debs vs rpms for linux release

cprice404:


bruuuuuuuce (approved):
:shipit: , I remember running into a similar issue when adding arm amis 😅 .

PR #208, opened 2022-12-14
chore: break up release action into two steps

cprice404:
The existing release action contains a step where it needs to commit a change to Cargo.toml.  This is important so that all of the other steps can build binaries that have their correct version number.

However, because all of the subsequent actions are run against the same git sha that triggered this action, they do not end up seeing that commit, so they use the previous commit with the old version number.

This breaks up the release logic into two phases: one phase that bumps the cargo version, tests the build, and commits the change to Cargo.toml; and then another phase that runs all of the remaining actions off of the new HEAD of the release branch.


cprice404:
this looks like a big diff but it is basically just splitting all of the logic into a second file; there are no changes in the new file.

pgautier404 (approved):
LGTM!

PR #210, opened 2022-12-16
chore: Try using SpectralOps rust release template.

schwern:
From https://github.com/SpectralOps/rust-ci-release-template
> This repo serves as a live template, and reference for building your own CI powered Rust release process on Github Action. This was built to fill a gap that Rust has, and Go doesn't have because Go has GoReleaser.

Giving this a try as a manual release process to see if it works and builds Mac ARM bottles and others. If it does, we can integrate it into the automated workflows.

### DIFF
--- a/.github/workflows/manual-release-trial.yml
+++ b/.github/workflows/manual-release-trial.yml
@@ -1,0 +1,179 @@
+# From https://github.com/SpectralOps/rust-ci-release-template
+# Giving this a try.
+name: Manual Build and Release
+on:
+  workflow_dispatch:
+
+env:
+  BIN_NAME: momento
+  PROJECT_NAME: momento-cli
+  REPO_NAME: momentohq/momento-cli
+  BREW_TAP: momentohq/homebrew-tap
+
+jobs:
+  dist:
+    name: Dist
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false # don't fail other jobs if one fails
+      matrix:
+        build: [x86_64-linux, aarch64-linux, x86_64-macos, x86_64-windows] #, x86_64-win-gnu, win32-msvc
+        include:
+        - build: x86_64-linux
+          os: ubuntu-20.04
+          rust: stable
+          target: x86_64-unknown-linux-gnu
+          cross: false
+        - build: aarch64-linux
+          os: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


cprice404:
do you think we could strip this down to just the mac parts and then try it out?  I've spent the better part of this week battling to get the linux stuff working, and stole a bunch of stuff from `bat` for it, and it's finally all working.  I also had to spin up about 10 different EC2 instances to test the debs and rpms on a bunch of different OS's.  I'd really like to avoid touching the linux parts now that they are all working :)

cprice404 (commented):
this looks great but would love to just trim it down to the mac parts.

schwern:
> this looks great but would love to just trim it down to the mac parts.

This is just a test to see if it works at all, not the final release script, so we might as well test it all.

Ultimately, my preference would be to have a single point for building everything, but we don't need to decide that now.

cprice404:
> > this looks great but would love to just trim it down to the mac parts.
> 
> This is just a test to see if it works at all, not the final release script, so we might as well test it all.

But if I'm reading the code right, if this works, it's going to attach a bunch more binaries to the current latest release, right?  And either overwrite the existing binaries that are attached to that release, or add a bunch more alongside them?

If you just want to test this then maybe we should create a bogus temporary release and modify the "tag" step in this one to just return that release version, hard-coded?

You can make the call on this; I'll give a 🚢 but if you run this and it modifies the current latest release, please do another run of the existing "manual release" job afterward so that it will produce a new "latest" release that has the current set of working artifacts on it.


> Ultimately, my preference would be to have a single point for building everything, but we don't need to decide that now.

I agree - I think we already have this, in `execute-release.yml`, and it is very similar to this.  I'd really prefer that we just add the missing mac parts into that when we know what we want them to look like so that we don't risk regressing on all the work that was done to get that action working.

cprice404 (approved):

cprice404:
right now our installation docs for linux include a hard-coded URL to an older release because we wanted to have a working installation method while we were fixing the linux builds.  but FYI i am planning on switching those docs back to recommend that users just grab the artifacts from the latest release, hopefully on Monday.  I will be switching the canary back to use the latest release at the same time.  So it's important that we don't break the linux/windows artifacts on the latest release if possible.

PR #215, opened 2022-12-21
chore: put momento binary at root of archive files

cprice404:
This commit changes the code that produces the archive files
so that the momento binary will be at the root.  It seems as
though homebrew formulae won't work properly for linux if the
binary is at a nested path.


bruuuuuuuce (approved):
:shipit:

PR #219, opened 2023-01-04
chore: sign windows installer

malandis:
We use a code signing certificate to sign the windows binary and windows installer. We also refactor the publish-windows-assets job steps to be more self contained.

### DIFF
--- a/.github/workflows/execute-release.yml
+++ b/.github/workflows/execute-release.yml
@@ -293,7 +293,7 @@ jobs:
           DISTRIBUTABLE_FILE_PREFIX: ${{ steps.build.outputs.distributable_file_prefix }}
         run: |
           $zipFilename = "$env:DISTRIBUTABLE_FILE_PREFIX.zip"
-          $zipPath = Get-Location + "\" + $zipFilename
+          $zipPath = ".\$zipFilename"
           Compress-Archive -LiteralPath $env:MOMENTO_BINARY_PATH -DestinationPath $zipPath
 
           echo "::set-output name=asset_path::$zipPath"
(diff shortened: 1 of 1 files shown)


.github/workflows/execute-release.yml
```diff
@@ -248,65 +248,123 @@ jobs:
...
+      - name: Delete PFX certificate
+        env:
+          CERT_PATH: ${{ steps.write_pfx.outputs.pfx_path }}
+        run: Remove-Item -Path $env:CERT_PATH
+
+      - name: Create zip and installer urls
+        id: create_release_urls
+        env:
+          ZIP_FILENAME: ${{ steps.create_zip.outputs.asset_name_zip }}
+          MSI_FILENAME: ${{ steps.create_zip.outputs.asset_name_msi }}
```
pgautier404, line 340:
```suggestion
          MSI_FILENAME: ${{ steps.build_installer.outputs.asset_name_msi }}
```

malandis: Thank you! Fixed in [36080c6](https://github.com/momentohq/momento-cli/pull/219/commits/36080c6875d3ea674748dc9a8e9786f87270c780)

pgautier404 (approved):
This is a tremendous improvement, thanks!

pgautier404 (dismissed):

cprice404 (commented):
This looks great to me but before we merge it, we need to think about what will happen if the signing fails (I believe we haven't been able to test it on github yet, right?)

If it fails after the github release is created, then the CLI canary will fail, which will cause problems in production.

So we need to be sure that either (a) if it fails, it will fail before the release is created, or (b) we have tested the signing code prior to merging this.

If you need some ideas on how to test the signing code without going through the full release process let me know, I can help.

malandis:
> This looks great to me but before we merge it, we need to think about what will happen if the signing fails (I believe we haven't been able to test it on github yet, right?)
> 
> If it fails after the github release is created, then the CLI canary will fail, which will cause problems in production.
> 
> So we need to be sure that either (a) if it fails, it will fail before the release is created, or (b) we have tested the signing code prior to merging this.
> 
> If you need some ideas on how to test the signing code without going through the full release process let me know, I can help.

@cprice404 the build passes here https://github.com/momentohq/momento-cli/actions/runs/3843526807/jobs/6545864685

pgautier404 (approved):

PR #221, opened 2023-01-06
chore: update installer instructions in README

cprice404:
Prior to this commit we had temporarily linked to a specific
version of the installers in the README, as we were rolling out
some changes to the release process that caused intermediate
breakage in the latest releases.

Now that the release process has stabilized, this commit updates
the docs back to guide users to the latest release.  We also
include more information about the new installer formats available
for linux and windows, and reduce some repetition in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,23 +9,56 @@ Command-line tool for managing Momento Serverless Cache.  Supports the following
 * Create, list, and delete Momento caches
 * Get and set values in a Momento cache
 
-## Quick Start
 
-Please refer to the installation instructions for Linux [here](https://github.com/momentohq/momento-cli/blob/main/README.md#linux) and Windows [here](https://github.com/momentohq/momento-cli/blob/main/README.md#windows).
+
+## Installation
+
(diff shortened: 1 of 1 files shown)


cprice404:
will file a corresponding PR for dev docs repo shortly

pgautier404 (approved):
This looks great!

NoSQLKnowHow:
Why is there an Upgrade section that only includes a brew commands and linux installs, but then there is another upgrade after the install section? Am I reading the changed file wrong?

cprice404:
> Why is there an Upgrade section that only includes a brew commands and linux installs, but then there is another upgrade after the install section? Am I reading the changed file wrong?

@NoSQLKnowHow here's a link to the rendered version, with the new content: https://github.com/momentohq/momento-cli/blob/7fc2247ec79d36bc4b35e1761835f722dbf0a02f/README.md

NoSQLKnowHow (approved):
That makes way more sense to see it like that. approved.

schwern (approved):

PR #229, opened 2023-01-24
feat: improve & refresh cli

kvcache:
* update clap command line parser, use the preferred macros
* update other libraries
* consolidate endpoint flags, make them not so position dependent
* fix all clippy violations
* add some help docs
* fix all clippy violations
* bring `momento` pr's to the normal Momento clippy low-bar threshold (still allowing expect by default)
  We did all that work a while back to add CliError, so I fixed all of the straggler call sites still using unwrap().

## command rework for more fluent consistency
An example of probably the most complex command we currently have:
```
momento cache set --help`
Store an item in a cache

Usage: momento cache set [OPTIONS] <KEY|--key <KEY>> <VALUE|--value <VALUE>>

Arguments:
  [KEY]    Cache key under which to store the value
  [VALUE]  Cache value to store under the key. This will be stored as UTF-8 bytes.

Options:
      --cache <CACHE>        Name of the cache you want to use. If not provided, your profile's default cache is used.
  -e, --endpoint <ENDPOINT>  An explicit hostname to use; for example, cell-us-east-1-1.prod.a.momentohq.com
      --verbose              Log more information
      --key <KEY>
  -p, --profile <PROFILE>    User profile [default: default]
      --value <VALUE>
      --ttl <TTL_SECONDS>    Max time, in seconds, that the item will be stored in cache
  -h, --help                 Print help
```

Now `momento` is aware of what the subcommands actually do, and expects the relevant arguments.
For example, `momento cache set some_key some_value` will do exactly what you would expect.

Additionally, `momento cache get some_key --key some_key` will refuse, since 2 key specs are
not allowed. Probably someone made a mistake and we don't want to do the wrong thing.

I've removed several short codes that seem to be more noise than value.

I added and fixed help text for a few commands as well.

Also I fixed the exit block. It would happily swallow the final stderr output unless you passed
it --verbose, which seems far too conservative a position to take with stderr.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,6 +10,7 @@ use log::{debug, error, LevelFilter};
 use utils::console::console_info;
 use utils::user::get_creds_and_config;
 
+use crate::utils::console::console_info;
 #[cfg(feature = "login")]
 use crate::utils::user::clobber_session_token;
 
@@ -238,23 +239,7 @@ enum CacheCommand {
     },
 }
 
-async fn entrypoint() -> Result<(), CliError> {
-    let args = Momento::parse();
-
-    let log_level = if args.verbose {
-        LevelFilter::Debug
-    } else {
-        LevelFilter::Off
-    }
-    .as_str();
-
-    env_logger::Builder::from_env(
-        Env::default()
-            .default_filter_or(log_level)
-            .default_write_style_or("always"),
-    )
-    .init();
-
+async fn run_momento_command(args: Momento) -> Result<(), CliError> {
     match args.command {
         Subcommand::Cache {
             endpoint,
@@ -397,13 +382,28 @@ async fn entrypoint() -> Result<(), CliError> {
 
 #[tokio::main]
 async fn main() {
+    let args = Momento::parse();
+
+    let log_level = if args.verbose {
+        LevelFilter::Debug
+    } else {
+        LevelFilter::Error
+    }
+    .as_str();
+
     panic::set_hook(Box::new(move |info| {
         error!("{}", info);
     }));
 
-    if let Err(e) = entrypoint().await {
-        // We should not exit nonzero without providing the message.
-        eprintln!("{}", e);
+    env_logger::Builder::from_env(
+        Env::default()
+            .default_filter_or(log_level)
+            .default_write_style_or("always"),
+    )
+    .init();
+
+    if let Err(e) = run_momento_command(args).await {
+        console_info!("{}", e);
         exit(1)
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,6 +6,7 @@ use commands::login::LoginMode;
 use env_logger::Env;
 use error::CliError;
 use log::{debug, error, LevelFilter};
+#[cfg(feature = "login")]
 use utils::console::console_info;
 use utils::user::get_creds_and_config;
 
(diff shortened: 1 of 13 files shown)


src/main.rs
```diff
@@ -33,159 +33,208 @@ struct Momento {
...
+        )]
         cache_name: Option<String>,
+
         // TODO: Add support for non-string key-value
-        #[structopt(long, short)]
-        key: String,
+        #[arg(help = "Cache key under which to store the value")]
+        key: Option<String>,
+        #[arg(long = "key", value_name = "KEY")]
+        key_flag: Option<String>,
```
cprice404, line 203: I am still kind of leaning towards removing these rather than supporting both unless someone gives us pushback?

src/main.rs
```diff
@@ -333,7 +402,8 @@ async fn main() {
     }));
 
     if let Err(e) = entrypoint().await {
-        console_info!("{}", e);
+        // We should not exit nonzero without providing the message.
+        eprintln!("{}", e);
```
cprice404, line 406: I have been trying to use the `console_` macros for all non-data output so that we have the option of doing things like also writing it to a log file etc without having to come change all the code.  any reason why we can't go through those macros here and need to go directly to eprintln?

cprice404 (commented):
this mostly seems good just a few questions/comments

src/main.rs
kvcache: Yes, it was eating output but I think I had a cached build and looking at it again it is in fact the panic hook above that is eating output. I'll update thanks.

src/main.rs
kvcache: idk it's not a battle i'll take up right now. It seems easier to get approval for additions rather than replacements, and though I can't relate, I can conceive of a person wanting to use the --flags. If you want the flags gone I'll take them away. You want to chat on slack about it and make a choice here?

kvcache: Per that epic slack thread, the consensus seems to be for supporting both due to external artifacts referencing the flags and a just reluctance to break them.

cprice404 (dismissed):
build is failing but otherwise 🚢

kvcache:
> build is failing but otherwise 🚢

It's because of fork action secrets permissions and the way we wrote integration tests:
```
thread 'tests::momento_configured_default_profile' panicked at 'Unexpected failure.
code-1
stderr=```"ERROR: Malformed Auth Token\n"```
command=`"/home/runner/work/momento-cli/momento-cli/target/debug/momento" "configure"`
stdin=`""`
code=1
stdout="Token: Default Cache [default-cache]: Default Ttl Seconds [600]: "
stderr="ERROR: Malformed Auth Token\n"
', /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/core/src/ops/function.rs:251:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

If it fails on push I'll fix it.

cprice404:
> > build is failing but otherwise 🚢
> 
> It's because of fork action secrets permissions and the way we wrote integration tests:
> 
> ```
> thread 'tests::momento_configured_default_profile' panicked at 'Unexpected failure.
> code-1
> stderr=```"ERROR: Malformed Auth Token\n"```
> command=`"/home/runner/work/momento-cli/momento-cli/target/debug/momento" "configure"`
> stdin=`""`
> code=1
> stdout="Token: Default Cache [default-cache]: Default Ttl Seconds [600]: "
> stderr="ERROR: Malformed Auth Token\n"
> ', /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/core/src/ops/function.rs:251:5
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> If it fails on push I'll fix it.

okay... might make other folks' lives easier if you did OSS PRs against the org repo until we have time to improve this.

cprice404 (approved):

Issue #236, opened 2023-01-26
On windows, CTRL-V does not work to paste token into terminal


cprice404:
According to the support team, this may be a known/common thing in Windows and Shift+Insert or Right-click are the preferred methods.

Need to confirm that those work, and then maybe add a MOTD or some text in the help panel to explain it.

cprice404:
tested in my windows environment.  shift+insert or right-click+paste seemed to work; I will add a help message to the configure command that suggests trying these.

PR #238, opened 2023-01-26
fix: prompt Windows users re: alternate mechanisms for pasting.

cprice404:


src/commands/configure/configure_cli.rs
```diff
@@ -81,6 +81,10 @@ async fn prompt_user_for_creds(profile_name: &str) -> Result<Credentials, CliErr
         .await
         .unwrap_or_default();
 
+    console_info!("Please paste your Momento auth token.  (If you do not have an auth token, use `momento account` to generate one.)");
+    console_info!("Windows users: if CTRL-V does not work, try right-click or SHIFT-INSERT to paste.");
+    console_info!("");
```
kvcache, line 86: consider only telling people if it's relevant?
```suggestion
    if cfg!(windows) {
        console_info!("If CTRL-V does not work, try right-click or SHIFT-INSERT to paste.");
        console_info!("");
    }
```

cprice404: the use case where this is causing problems for users is in the instruqt environment, where the momento binary is running on ubuntu but the user's machine is windows and they are running in a browser.

So if I accept this change those users still won't see the message, right?

cprice404:
Relates to https://github.com/momentohq/momento-cli/issues/236

src/commands/configure/configure_cli.rs
cprice404: I could theoretically key this on some environment variable like MOMENTO_INSTRUQT_ENVIRONMENT=true but I wasn't sure it was worth it.  LMK if you feel like that's necessary.

kvcache: i sure don't think that is worth it no

cprice404: okay.  the other option would be to not put this text here and instead put it in the right-nav on the instruqt project but I felt like it would be more likely for users to miss it there.

@eaddingtonwhite lmk if you have any opinions on this.

cprice404: i'd like to get this one rolled out today if possible, one way or the other.  happy to make changes though.

kvcache (approved):

src/commands/configure/configure_cli.rs
eaddingtonwhite: I think this is good for now agree. Lets keep brain storming can fast follow if we can make this better.

Issue #235, opened 2023-01-26
Improve message about default cache


cprice404:
when configuring for a second time, this message is printed to the console:

```
default-cache as the default already exists
```

need to make this more clear or omit it.

PR #239, opened 2023-01-27
fix: don't log confusing message if the user runs configure a second time

cprice404:
Prior to this commit if the user ran configure more than once, on the second and subsequent attempts they would get a confusing message printed about the default cache already existing.  This commit removes that message as it doesn't provide any real value to the user.


cprice404:
relates to # https://github.com/momentohq/momento-cli/issues/235

eaddingtonwhite (dismissed):

cprice404:
rebased

kvcache (approved):
Yeah I agree

PR #240, opened 2023-01-27
chore: update windows install instructions for winget

malandis:
We add winget as the primary installation path for Windows users. Previously the only
installation path was by downloading the installer/zip from the GitHub
release page. We demote this to an alternative path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,14 @@ which you may add anywhere you like in your execution path.
 
 ### Windows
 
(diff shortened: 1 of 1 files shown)


malandis:
TODO: we will have to translate this to Nihongo and Português

poppoerika (dismissed):

cprice404 (approved):
We need to make the corresponding change in public-dev-docs too please :)

PR #242, opened 2023-01-27
chore: add some initial basic tests for config/creds ini interactions

cprice404:
This commit just adds some early unit tests for the ini_config
module.  A minor refactor was required for some separation of
concerns so that the tests could be written without writing
files to disk.

In subsequent commits the functions in the ini_config will be
refactored further to make them more testable, but this commit
gets the tests in place so that we can hopefully ensure that
we don't break any functionality in the subsequent refactor.

The end goal here is to fix some issues that users reported
where empty tokens don't get updated properly, or leading/trailing
whitespace entered in the CLI input cause corruption to the
files; this will be addressed in later commits once the code
can be tested more easily.

### DIFF
--- a/src/commands/configure/configure_cli.rs
+++ b/src/commands/configure/configure_cli.rs
@@ -12,7 +12,7 @@ use crate::{
             open_file, prompt_user_for_input, read_file_contents, write_to_file,
         },
         ini_config::{
-            add_new_profile_to_config, add_new_profile_to_credentials, update_profile_values,
+            create_new_config_profile, create_new_credentials_profile, update_profile_values,
         },
         user::{get_config_for_profile, get_creds_for_profile},
     },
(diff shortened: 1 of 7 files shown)


cprice404:
For best results you will want to review this with whitespace changes hidden, and you will need to expand the "large diffs are hidden" on the `configure_cli.rs` file because it is where most of the important changes are :)

poppoerika (approved):
Thank you for cleaning up the mess I've created with modifying files!
Looks good to me. As you mentioned in the PR, getting line numbers for existing profiles is fragile so thank you for pointing out that too.

src/commands/configure/configure_cli.rs
```diff
@@ -223,104 +222,99 @@ async fn set_file_read_write(path: &str) -> Result<(), CliError> {
...
-                                existing_profile_starting_line_num,
-                                file_contents.clone(),
-                                FileTypes::Config(cf),
-                            ) {
-                                Ok(v) => v,
-                                Err(e) => return Err(e),
-                            }
+        trimmed.split('\n').map(|line| line.to_string()).collect()
+    }
+}
```
kvcache, line 258: if we really want to do this maybe something like this would be better
```rust
fn trim_file_contents(lines: Vec<String>) -> Vec<String> {
    lines.into_iter()
        .skip_while(String::is_empty)
        .collect::<Vec<String>>() // skip iterators don't support double-ended so let's cheat
        .into_iter()
        .rev()
        .skip_while(String::is_empty)
        .collect::<Vec<String>>()
        .into_iter()
        .rev()
        .collect()
}
```

src/commands/configure/configure_cli.rs
```diff
@@ -360,68 +354,146 @@ fn find_existing_profile_start(file_contents: Vec<String>, profile_name: &str) -
...
-            push_to_file_contents(file_contents, FileTypes::Config(cf), profile_name)
-        }
-    };
-    updated_file_contents
+    fn test_content_to_lines(file_contents: &str) -> Vec<String> {
+        file_contents
+            .trim()
+            .split('\n')
+            .map(|line| line.to_string())
+            .collect()
```
kvcache, line 395: seems interesting that we would take a lines-oriented api like `next_line()` and wrap it as a vec of strings with trailing newlines, and the test function here also seems interesting as a result. Not something to change in this pr but we should take a fresh look at how we marshal these files with the benefit of more experience over the last year.

src/utils/ini_config.rs
```diff
@@ -182,30 +164,286 @@ fn replace_value(
...
+        ",
+        );
+        assert_eq!(expected_text.trim(), profile_text)
+    }
+
+    #[test]
+    fn update_profile_values_credentials_one_existing_profile() {
+        // TODO
+        // TODO these line number bits are fragile and we should refactor to prevent them from
+        // being necessary.  Will hit in a subsequent PR.
```
kvcache, line 248: okay yeah, gotcha you're already thinking about the above

kvcache (approved):
looks like a fine improvement. I have a suggestion for the trim function, but I'm not super certain it makes a lot of sense to even do it that way (e.g., since we have a "read file contents" explicit function why not do it in there)

PR #245, opened 2023-01-29
chore: simplify signature for ini update_profile_values

cprice404:
This commit contains no functional changes, just a refactor. Prior to this commit there was some ini-specific logic for finding profiles defined in the configure_cli mod.  This commit moves it to the ini_config mod so we can try to keep all of the ini-specific logic there.

In addition, the `update_profile_values` fn in ini_config required you to pass in information about all of the profile line numbers as arguments; this made it very fragile to write tests, because you either needed to call a separate function to determine those line numbers or maintain the test line numbers by hand.  This commit encapsulates the line number logic inside of the function, so that callers don't have to worry about it and tests are less fragile.

### DIFF
--- a/src/commands/configure/configure_cli.rs
+++ b/src/commands/configure/configure_cli.rs
@@ -12,7 +12,8 @@ use crate::{
             open_file, prompt_user_for_input, read_file_contents, write_to_file,
         },
         ini_config::{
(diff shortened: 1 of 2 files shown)


kvcache (approved):
looks fine

PR #246, opened 2023-01-29
chore: refactor ini signatures to accept &str to reduce cloning

cprice404:
This commit changes the signatures of several functions in the ini_config module to accept slices of &str or String, rather than just Vec<String>, so that we don't need to do so much cloning before calling these functions.  Also removes calls to clone in several call sites.

### DIFF
--- a/src/utils/ini_config.rs
+++ b/src/utils/ini_config.rs
@@ -20,9 +20,9 @@ pub fn create_new_config_profile(profile_name: &str, config: Config) -> Vec<Stri
     ]
 }
 
-pub fn update_profile_values<T: AsRef<str>>(
+pub fn update_profile_values(
     profile_name: &str,
-    file_contents: &[T],
+    file_contents: &[impl AsRef<str>],
     file_types: FileTypes,
 ) -> Result<Vec<String>, CliError> {
     let existing_profile_line_numbers = match find_profile_line_numbers(file_contents) {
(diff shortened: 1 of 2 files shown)


src/utils/ini_config.rs
```diff
@@ -20,12 +20,12 @@ pub fn create_new_config_profile(profile_name: &str, config: Config) -> Vec<Stri
     ]
 }
 
-pub fn update_profile_values(
+pub fn update_profile_values<T: AsRef<str>>(
     profile_name: &str,
-    file_contents: Vec<String>,
+    file_contents: &[T],
```
kvcache, line 25: did you consider not making the function explicitly generic over a bound type parameter, but instead using the equivalent parameter impl syntax? Imo it reads more nicely with the type in 1 place rather than 2, and keeps the relevant information closer to where it's applicable (i.e., it does not matter that the function _is_ generic over a trait parameter; it only matters that when you're passing file_contents it needs to be a slice of something that can look like an &str)
```suggestion
pub fn update_profile_values(
    profile_name: &str,
    file_contents: &[impl AsRef<str>],
```
(here and everywhere)

cprice404: nope i didn't know that was a thing.  i agree it's better if it works, will try it.

kvcache: Context for other readers who may be curious about this:

The `impl SomeTrait` thing is just some syntactic sugar that means exactly the same thing as your original change. It's a nicer way to express the mundane kind of variance without all the configuration the template parameters provide. If you needed to add a `where T: Something + Something2 + Something3` fancy parameter then you couldn't use `impl SomeTrait` (well, without defining that combination as its own trait, which is sometimes still the right thing to do). It's a nice tip to readers that the intent is purely "simple type compatibility, _and nothing else_"

cprice404: done in https://github.com/momentohq/momento-cli/pull/246/commits/5822d6bf1de903fc3c64fead4067cf795f9b4f19

cprice404:
This one has been rebased and then I added a commit that implements Kenny's suggestion.

kvcache (approved):
nice

PR #247, opened 2023-01-29
fix: add tests to cover case with empty token in config file, fix bug

cprice404:
A user reported that if the credentials file has an empty token (e.g. if they pressed enter without pasting a token in when running the configure command), subsequent calls to the configure command did not update the file properly.

This commit adds test coverage for this case and fixes the bug that was preventing the new value from being written.


kvcache (dismissed):
> This commit adds test coverage for this case and fixes the bug that was preventing the new value from being written.

exemplary.

PR #248, opened 2023-01-29
chore: clean up integration tests, support MOMENTO_CONFIG_DIR

cprice404:
This commit contains a smattering of tweaks to try to make the integration tests more sustainable:

1. Introduces support for a MOMENTO_CONFIG_DIR env var, which, if provided, will be used instead of `~/.momento`.
2. Use MOMENTO_CONFIG_DIR env var (pointing to a temp dir) in all of the integration tests, so that developers' ~/.momento files don't get overwritten when running the tests.
3. Remove a bunch of env vars that used to be required for the tests; now you only need TEST_AUTH_TOKEN and the rest of the values are created by the test code as needed.
4. Remove `configure_profiles_test` as it was almost entirely redundant to the `momento_default_profile` test.
5. Each integration test module now creates its own momento config temp dir at startup, to ensure that the config files are not polluted by previous tests.
6. All necessary config file setup is now handled inside of the integration tests rather than in the github workflow yaml.
7. Adds coverage of the new positional arguments for `momento set` and `momento get`.
8. Adds coverage of the new positional arguments for `momento cache create` and `momento cache delete`.
9. Adds coverage of the new alternative `--cache` flag for `momento cache create` and `momento cache delete`.


kvcache:
> Adds coverage of the new positional arguments for

thank you for doing this @cprice404

tests/common/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    println!();
+    let final_creds_path = config_dir.join("credentials");
+    debug_output_file(&final_creds_path);
+}
+
+fn debug_output_file(file_path: &PathBuf) {
+    println!("Momento config / creds file ({file_path:?}) contents:");
+    let lines = io::BufReader::new(
+        File::open(file_path).unwrap_or_else(|_| panic!("Unable to open file: {file_path:?}")),
```
kvcache, line 56:
```suggestion
        File::open(file_path).unwrap_or_else(|e| panic!("Unable to open file: {file_path:?}, {e:?}")),
```

kvcache (approved):
l's gtm

cprice404:
@kvcache I had to rebase this and then fix a bunch of gh workflow things for how the tests get launched

kvcache (approved):

.github/workflows/on-pull-request.yml
```diff
@@ -52,34 +48,7 @@ jobs:
...
-        run: |
-          cat <<EOF > ~/.momento/config 
-          [default]
-          cache=$TEST_CACHE_DEFAULT
-          ttl=900
-          [$TEST_PROFILE]
-          cache=$TEST_CACHE_WITH_PROFILE
-          ttl=1000
-          EOF
-          cat ~/.momento/config
```
poppoerika, line 82: Woohoo

poppoerika (approved):

poppoerika:
With this change, we want to update `CONTRIBUTING.md` https://github.com/momentohq/momento-cli/blob/main/CONTRIBUTING.md so that users don't need to be warned about running the integration tests unnecessarily 😄

PR #249, opened 2023-01-30
chore: simplify logic for updating config/credentials files

cprice404:
This PR contains a series of commits that make simplifications to the code for updating the config/credentials files.  Should be easier to maintain as a result, as well as significantly reducing the number of copy operations that are being made when updating the files.


kvcache (dismissed):
This looks reasonable to me

cprice404:
rebased

PR #257, opened 2023-02-03
chore: update sdk

kvcache:
consume the latest sdk with the new error types. This also lets
`topic subscribe` do a little bit better with output when the
connection is interrupted without an error - whether due to
inactivity or a service deployment and h2 goaway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /target
 .idea/
+.vscode/
 client-cli.iml
 .DS_Store
 .vs/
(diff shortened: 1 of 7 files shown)


cprice404 (approved):

mjindalmotiv8:
Just one question - is there any way to say Unsupported Operation - for example, I had tried to use topics on an unsupported memcache

kvcache:
> Just one question - is there any way to say Unsupported Operation

we'd have to add that support on the service, which we do not currently have.

PR #258, opened 2023-02-15
chore: introduce separate lib and app for binary vs. cli options

cprice404:
We have had a customer request for tab completion in the CLI.  In
order to support that we need to move the CLI options into their
own library that can be referenced via build.rs from the main app.

This commit introduces the new directory structure.  No files are
actually moved yet.


kvcache (commented):
build is failing, but generally this looks fine

cprice404:
fixed the build.

kvcache (approved):

PR #260, opened 2023-02-16
chore: move CLI opts into library package

cprice404:
This moves the definition of the CLI options into the new
momento-cli-ops library package.  This will allow us,
in subsequent commits, to update the build.rs for the
binary to generate completion scripts for various shells.


momento-cli-opts/Cargo.toml
```diff
@@ -2,3 +2,7 @@
 name = "momento-cli-opts"
 version = "0.1.0"
 edition = "2021"
+
+[dependencies.clap]
+version = "4.1"
+features = [ "derive",]
```
kvcache, line 8: can this be removed from the other project?

momento/src/commands/login.rs
```diff
@@ -1,65 +1,65 @@
...
-                })),
-            }
-        }
-        momento::auth::LoginAction::ShowMessage(message) => {
-            console_info!("{}", message.text);
-            None
-        }
-    }
-}
+// use crate::utils::console::console_info;
```
kvcache, line 1: did you forget to finish this?

cprice404: sorry, I meant to add some comments about this.  I can't figure out how to do this `feature` stuff across the library vs app projects, so I decided to just comment out the login stuff for now, especially since folks should be picking that back up as a focal point pretty soon.  Do you have any objections to this?

cc: @rlinehan @bruuuuuuuce

momento-cli-opts/Cargo.toml
cprice404: good call, will check

cprice404: nope, there is a `parse` function that we use in `main` that comes from this import.

momento/src/commands/login.rs
bruuuuuuuce: I don't have enough context rn as to where we are with login/signup flow to have an opinion on this rn, but is good to know that its being commented out for this change.

kvcache: instead of commenting this out you could also just remove cfg() and use `#[command(hide = true)]`? That would make sure you're not making any changes here that unexpectedly break the feature's compilation.

cprice404: okay.  but FWIW i'm pretty opposed to us doing stuff this way going forward.  we've been maintaining the compilation of this code for like, 6 months now without anyone actually using it or working on it.  next time we should just delete it and pull it back out of the git history when we are actually going to use it.

cprice404: and now i need to go update CI jobs and a bunch of other stuff to get rid of the features, and this is all orthogonal to the task I have actually been trying to work on :)

kvcache: I agree - we should have deleted the code instead of cfg()ing it.

cprice404: fixed in https://github.com/momentohq/momento-cli/pull/260/commits/9cff010d8e9b22a911bc8a1e781a0f424bfd8e9d

kvcache (approved):

PR #261, opened 2023-02-16
chore: generate zsh/bash completion scripts

cprice404:
In a subsequent commit we will need to package these in the appropriate places in the releases in order for them to work.


kvcache (dismissed):

cprice404:
rebased to make sure CI still passes

PR #266, opened 2023-02-22
feat: add support for flushing a cache

cprice404:


mjindalmotiv8 (approved):
Thank you

anp13 (approved):

momento-cli-opts/src/lib.rs
```diff
@@ -210,6 +210,12 @@ pub enum CacheCommand {
     #[command(about = "List all caches")]
     List {},
 
+    #[command(about = "Flush all contents from a cache")]
+    Flush {
+        #[arg(help = "Name of the cache to flush.", value_name = "CACHE")]
+        cache_name: String,
+    },
```
kvcache, line 217: so it looks like this would be `momento cache flush my_cache`? I think that's nice.

kvcache (approved):

PR #268, opened 2023-02-22
feat: hide signing key command from help, improve preview text

cprice404:
This commit moves the signing key command to "preview" status,
so that it doesn't show up in the default help.  It will still
show up in tab completion at the root of the command, but
is now prefixed with our "**PREVIEW**" label, and includes some
information in the subcommand help text about its preview status.

### DIFF
--- a/momento-cli-opts/src/lib.rs
+++ b/momento-cli-opts/src/lib.rs
@@ -93,10 +93,10 @@ To delete a topic, stop subscribing to it."
         hide = true,
         about = "**PREVIEW** Manage signing keys",
         before_help = "
-!!                                                              !!
-!!                        Preview feature                       !!
-!!   For more information, contact us at support@momento.com.   !!
-!!                                                              !!
+!!                                                                !!
+!!                        Preview feature                         !!
+!!   For more information, contact us at support@gomomento.com.   !!
+!!                                                                !!
 
 Signing keys can be used to generate pre-signed URLs that allow access to a value in the cache
 for a specified period of time.  The URLs can be distributed to browsers or devices to allow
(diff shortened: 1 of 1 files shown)


momento-cli-opts/src/lib.rs
```diff
@@ -89,7 +89,20 @@ To delete a topic, stop subscribing to it."
...
         operation: AccountCommand,
     },
-    #[command(about = "Manage signing keys")]
+    #[command(
+        hide = true,
+        about = "**PREVIEW** Manage signing keys",
+        before_help = "
+!!                                                              !!
+!!                        Preview feature                       !!
+!!   For more information, contact us at support@momento.com.   !!
```
danielamiao, line 98: i wish we had `momento.com` lol but we don't

momento-cli-opts/src/lib.rs
```diff
@@ -103,8 +116,17 @@ To delete a topic, stop subscribing to it."
         operation: SigningKeyCommand,
     },
     #[command(
+        hide = true,
         about = "**PREVIEW** Log in to manage your Momento account",
-        hide = true
+        before_help = "
+!!                                                              !!
+!!                        Preview feature                       !!
+!!   For more information, contact us at support@momento.com.   !!
```
danielamiao, line 124: same comment re: email domain

cprice404: fixed in https://github.com/momentohq/momento-cli/pull/268/commits/0167243bb71a3f2dff53a42fc27be4ba2bf378f7

momento-cli-opts/src/lib.rs
cprice404: fixed in https://github.com/momentohq/momento-cli/pull/268/commits/0167243bb71a3f2dff53a42fc27be4ba2bf378f7

danielamiao (approved):

PR #282, opened 2023-04-18
chore: Deprecating us-east1 as us-east4 is now setup.

mjindalmotiv8:
https://github.com/momentohq/core-infrastructure/issues/512

### DIFF
--- a/momento-cli-opts/src/lib.rs
+++ b/momento-cli-opts/src/lib.rs
@@ -212,11 +212,7 @@ pub enum CloudSignupCommand {
     Gcp {
         #[arg(long, short)]
         email: String,
-        #[arg(
-            long,
-            short,
-            value_name = "us-east4, us-central1, asia-northeast1"
-        )]
+        #[arg(long, short, value_name = "us-east4, us-central1, asia-northeast1")]
         region: String,
     },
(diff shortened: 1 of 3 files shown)


momento-cli-opts/src/lib.rs
```diff
@@ -212,11 +212,7 @@ pub enum CloudSignupCommand {
     Gcp {
         #[arg(long, short)]
         email: String,
-        #[arg(
-            long,
-            short,
-            value_name = "us-east1, us-east4, us-central1, asia-northeast1"
```
cprice404, line 218: not a big deal but anyone who doesn't upgrade their CLI is gonna still see this in the help.  Would be good to know what the error message looks like if they try to use that region with an older version of the CLI, and/or think about whether there's another way we want to surface these long term to prevent this issue.

just food for thought, no changes to this PR!

cprice404 (approved):

virratanasangpunth (approved):

momento-cli-opts/src/lib.rs
mjindalmotiv8: I have also checked in https://github.com/momentohq/momento-account-registry-data/pull/101 change - this should make it so that the signup lambda returns the region not supported error

PR #292, opened 2023-06-28
teardown gcp us east4

nidhidhamnani:
- chore: remove gcp asia-northeast1 from README
- chore: teardown us-east4


rlinehan:
@nidhidhamnani yeah, we did. I think what we need to do is get the translation updated in this readme. Let me check with some folks about that.

nidhidhamnani:
> @nidhidhamnani yeah, we did. I think what we need to do is get the translation updated in this readme. Let me check with some folks about that.

Thank you :)

PR #296, opened 2023-08-07
feat: add support for deleting an item from cli

pratik151192:
Added support to delete an item for a cache. Since we already had ```delete``` as a cache subcommand to delete the cache, I went with ```delete-item``` for this one. Let me know if you have better suggestions.

How was this change tested?

- Manually ran ```./target/debug/momento cache delete-item``` for a key I had inserted and verified it got deleted from the cache. A non-existent key doesn't throw an error.
- Added a test case for command verification similar to other commands.

### DIFF
--- a/momento/tests/momento_additional_profile_test.rs
+++ b/momento/tests/momento_additional_profile_test.rs
@@ -73,9 +73,16 @@ mod tests {
 
     async fn momento_cache_delete_key_with_profile(profile_name: &str) {
         let mut cmd = Command::cargo_bin("momento").unwrap();
-        cmd.args(["cache", "delete-item", "--key", "key", "--profile", profile_name])
-            .assert()
-            .success();
+        cmd.args([
+            "cache",
+            "delete-item",
+            "--key",
+            "key",
+            "--profile",
+            profile_name,
+        ])
+        .assert()
+        .success();
     }
 
     async fn momento_cache_list_with_profile(profile_name: &str, cache_name: &str) {
--- a/momento/tests/momento_additional_profile_test.rs
+++ b/momento/tests/momento_additional_profile_test.rs
@@ -71,6 +71,13 @@ mod tests {
             .stdout("value\n");
     }
 
+    async fn momento_cache_delete_key_with_profile(profile_name: &str) {
+        let mut cmd = Command::cargo_bin("momento").unwrap();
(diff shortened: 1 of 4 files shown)


cprice404:
@kvcache @eaddingtonwhite @danielamiao are you all okay with `momento cache delete-item` as the name for this subcommand?

momento/tests/momento_additional_profile_test.rs
```diff
@@ -71,6 +71,20 @@ mod tests {
             .stdout("value\n");
     }
 
+    async fn momento_cache_delete_key_with_profile(profile_name: &str) {
```
cprice404, line 74: just curious if this is the only place where there are tests for the other operations?

cprice404 (approved):
:+1: pending a bit of confirmation from other folks about the subcommand name.

momento/tests/momento_additional_profile_test.rs
pratik151192: yep this is the only one I saw and seemed relevant

eaddingtonwhite:
delete-item makes sense to me I think. 👍

danielamiao:
I'm OK with it, it's a bit too long for my liking but I don't have a better suggestion

kvcache:
* invalidate
* remove

did you consider these single word command names?

pratik151192:
@kvcache I did consider ```remove```. I preferred the word ```delete``` in the name as all our SDKs and API references use that word to get rid of a key. I am not opposed to ```remove``` though. @cprice404 thoughts?

Invalidate is something I didn't consider, however, it seems more of a policy than an operation. Invalidate can also indicate ```upsert``` and not necessarily a ```delete```.

cprice404:
i vote for consistency with the SDKs over brevity, especially since we have tab completion. so i'd rather not introduce a new term for now?

I think probably in the future before we `1.0` the CLI we should reorganize the top-level subcommands so that we have something like `momento cache create` and `momento item set`, so all of the item commands are in their own namespace and it's not confusing between the cache control plane operations. at that point in time it can be just `momento item delete`. but for now I think it would be confusing to use a word other than `delete` given that that is what it's called in the SDKs.

@kvcache can you live with `delete-item` for now?

kvcache:
I'm not going to block this pr. I disagree with the name but it's not a big deal: The one you want is not ambiguous or confusing, or even wrong.

PR #298, opened 2024-03-25
feat: Add rate limiting function

nand4011:
Add a function that executes another function with a rate limit. Will be used in the cloud linter.

Update the clap dependency to try and fix an error about using a nightly feature.

### DIFF
--- a/momento/src/commands/cloud_linter/linter_cli.rs
+++ b/momento/src/commands/cloud_linter/linter_cli.rs
@@ -1,0 +1,73 @@
+use std::future::Future;
+use std::sync::Arc;
+use std::time::Duration;
+
+use governor::DefaultDirectRateLimiter;
+
+#[allow(dead_code)] // remove after this is used outside a test
+async fn rate_limit<F, Fut, T>(func: F, limiter: Arc<DefaultDirectRateLimiter>) -> T
+    where
+        F: Fn() -> Fut,
+        Fut: Future<Output=T>,
+{
+    loop {
+        let permit = limiter.check();
+        match permit {
+            Ok(_) => {
+                return func().await;
+            }
+            Err(_) => {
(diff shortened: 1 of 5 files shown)


momento/src/commands/cloud_linter/linter_cli.rs
```diff
@@ -0,0 +1,73 @@
...
+                async move {
+                    let mut count = counter.lock().await;
+                    *count += 1;
+                }
+            }
+        };
+        let start_time = tokio::time::Instant::now();
+
+        let mut tasks = Vec::new();
+        for _ in 0..20 {
```
bruuuuuuuce, line 54: I think this test is testing that all 20 tasks are invoked, but not that they are rate limited in any way. I don't want to block on this, but just wanted to call that out if that was the intention

momento/src/commands/cloud_linter/linter_cli.rs
```diff
@@ -0,0 +1,73 @@
...
+        Fut: Future<Output=T>,
+{
+    loop {
+        let permit = limiter.check();
+        match permit {
+            Ok(_) => {
+                return func().await;
+            }
+            Err(_) => {
+                tokio::time::sleep(Duration::from_millis(100)).await;
```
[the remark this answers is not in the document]
nand4011: I don't think we need to do anything fancy since since the rate limit is so low.

momento/src/commands/cloud_linter/linter_cli.rs
nand4011: It checks that the tasks have all completed and that the duration of the tasks is more than one second. If there was no rate limit, it would be faster.

bruuuuuuuce (approved):
:shipit:

PR #303, opened 2024-04-01
feat: fetch metrics for the dynamo and elasticache resources

nand4011:
Change the get metadata functions for dynamo and elasticache to return a resource that will later be serialized into json.

Add code for fetching metrics based on the dynamo and elasticache resources.

Change the linter cli to print out the full json including the metrics.

### DIFF
--- a/momento/src/commands/cloud_linter/dynamodb.rs
+++ b/momento/src/commands/cloud_linter/dynamodb.rs
@@ -1,47 +1,184 @@
+use std::collections::HashMap;
 use std::sync::Arc;
 
 use aws_config::SdkConfig;
 use aws_sdk_dynamodb::types::{TimeToLiveDescription, TimeToLiveStatus};
 use governor::DefaultDirectRateLimiter;
+use phf::{phf_map, Map};
 use serde::{Deserialize, Serialize};
 
+use crate::commands::cloud_linter::metrics::{Metric, MetricTarget, ResourceWithMetrics};
+use crate::commands::cloud_linter::resource::{DynamoDbResource, Resource, ResourceType};
 use crate::commands::cloud_linter::utils::rate_limit;
 use crate::error::CliError;
 use crate::utils::console::console_info;
 
-#[derive(Serialize, Deserialize)]
+const DDB_TABLE_METRICS: Map<&'static str, &'static [&'static str]> = phf_map! {
+        "Sum" => &[
+            "ConsumedReadCapacityUnits",
+            "ConsumedWriteCapacityUnits",
+            "ReadThrottleEvents",
+            "WriteThrottleEvents",
+            "TimeToLiveDeletedItemCount",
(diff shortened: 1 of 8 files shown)


momento/src/commands/cloud_linter/dynamodb.rs
```diff
@@ -1,47 +1,184 @@
...
+        "Sum" => &[
+            "ConsumedReadCapacityUnits",
+            "ConsumedWriteCapacityUnits",
+            "ReadThrottleEvents",
+            "WriteThrottleEvents",
+            "TimeToLiveDeletedItemCount",
+            "TransactionConflict",
+            "ConditionalCheckFailedRequests",
+        ],
+        "Average" => &[
```
bruuuuuuuce, line 26: I think we also need the average of 

```
ConsumedReadCapacityUnits
ConsumedWriteCapacityUnits
```

I think I forgot to update the doc for these, I will do that now and make sure its updated

momento/src/commands/cloud_linter/dynamodb.rs
```diff
@@ -1,47 +1,184 @@
...
+};
+
+const DDB_GSI_METRICS: Map<&'static str, &'static [&'static str]> = phf_map! {
+    "Sum" => &[
+            "ConsumedReadCapacityUnits",
+            "ConsumedWriteCapacityUnits",
+            "ReadThrottleEvents",
+            "WriteThrottleEvents",
+        ],
+    "Average" => &[
```
bruuuuuuuce, line 47: same here

momento/src/commands/cloud_linter/metrics.rs
```diff
@@ -0,0 +1,156 @@
...
+    async fn append_metrics(
+        &mut self,
+        metrics_client: &Client,
+        limiter: Arc<DefaultDirectRateLimiter>,
+    ) -> Result<(), CliError> {
+        let metric_target = self.create_metric_target()?;
+        let metrics =
+            query_metrics_for_target(&metrics_client, Arc::clone(&limiter), metric_target).await?;
+        self.set_metrics(metrics);
+        self.set_metric_period_seconds(60 * 60 * 24);
```
bruuuuuuuce, line 60: This is 1 day right?

bruuuuuuuce (commented):
Looks good! Left a few clarifying questions/nits

momento/src/commands/cloud_linter/metrics.rs
nand4011: It is.

momento/src/commands/cloud_linter/dynamodb.rs
bruuuuuuuce: @nand4011 we want both the sum and the average for these metrics

bruuuuuuuce (approved):
:shipit:

PR #305, opened 2024-04-03
feat: write linter output to a gzipped JSON file

nand4011:
Add a step to check if the output file will be writable before making any network calls.

Gzip compress the final json and write it to linter_results.json.gz in the directory the linter command is executed from.

Check the AWS credentials before starting to provide a better error message than would occur if we made an AWS call with invalid credentials.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "adler"
+version = "1.0.2"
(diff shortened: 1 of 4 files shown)


bruuuuuuuce (dismissed):
:shipit:

bruuuuuuuce (approved):
:shipit:

bruuuuuuuce (approved):
:shipit:

PR #306, opened 2024-04-11
fix: speed up cloud linter command

bruuuuuuuce:
This command just speeds up the elasticache/getMetrics api calls. We need to refactor a bit how the limiter is working. We should be able to query `DescribeDynamodbTables` at a higher rate that we query the `DescribeTimeToLive` api. I spent a bit of time trying to configure separate rate limiters for the ddb api calls, but it was not behaving as expected. I think we might need to first describe all of the ddb tables, maybe have some heuristics around which tables are worth attempting to get a ttl off of, and then call those with a separate slower rate limiter

Before
Screen Shot 2024-04-10 at 6 15 17 PM

After
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ client-cli.iml
 bin/
 momento.exe
 obj/
+*.gz
+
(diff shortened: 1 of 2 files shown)


bruuuuuuuce:
I'm working on adding a heuristic around querying the time to live metric, which will supersede this pr

bruuuuuuuce:
closing in favor of https://github.com/momentohq/momento-cli/pull/307

PR #307, opened 2024-04-11
fix: use heuristics to determine when to query ddb table ttl

bruuuuuuuce:
The describe_time_to_live api call is heavily throttled, we don't want to have to call it for every dynamodb table. This pr separates out the describe ttl call into a separate function, and only calls it on tables who have at least an average of 100tps over the course of a month. We can tune this heuristic later on

Before
Screen Shot 2024-04-11 at 11 49 41 AM

After
Screen Shot 2024-04-11 at 11 36 38 AM

### DIFF
--- a/momento/src/commands/cloud_linter/dynamodb.rs
+++ b/momento/src/commands/cloud_linter/dynamodb.rs
@@ -164,33 +164,91 @@ impl ResourceWithMetrics for DynamoDbResource {
     }
 }
 
+pub(crate) async fn append_ttl_to_appropriate_resources(
+    config: &SdkConfig,
+    mut resources: Vec<Resource>,
+    limiter: Arc<DefaultDirectRateLimiter>,
+) -> Result<Vec<Resource>, CliError> {
+    log::debug!("describing ttl for dynamodb tables");
+    let ddb_client = aws_sdk_dynamodb::Client::new(config);
+    let describe_ddb_ttl_bar =
+        ProgressBar::new(resources.len() as u64).with_message("Describing Dynamo DB Ttl");
+    describe_ddb_ttl_bar
+        .set_style(ProgressStyle::with_template("  {msg} {bar} {eta}").expect("invalid template"));
+    for resource in &mut resources {
+        describe_ddb_ttl_bar.inc(1);
+        match resource {
+            Resource::DynamoDb(r) => {
+                if r.resource_type == ResourceType::DynamoDbGsi {
+                    continue;
+                }
+                let consumed_write_ops_index = r.metrics.iter().position(|p| {
+                    p.name
+                        .eq_ignore_ascii_case("consumedwritecapacityunits_sum")
+                });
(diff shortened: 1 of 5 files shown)


momento/src/commands/cloud_linter/dynamodb.rs
```diff
@@ -164,33 +164,91 @@ impl ResourceWithMetrics for DynamoDbResource {
...
+                        .eq_ignore_ascii_case("consumedwritecapacityunits_sum")
+                });
+                match consumed_write_ops_index {
+                    Some(index) => {
+                        let consumed_write_capacity =
+                            r.metrics.get(index).expect("index should exist");
+                        let sum: f64 = consumed_write_capacity.values.iter().sum();
+                        // a basic heuristic around whether or not we care to check to see if a ttl exists on a ddb table. If the dynamodb table
+                        // has less than 100 tps average, then we don't care to check if ttl is enabled or not.
+                        if sum < 100.0 * 60.0 * 60.0 * 24.0 * 30.0 {
```
cprice404, line 196: I'd probably check in with @eaddingtonwhite on this; my gut is that 100 might be a bit too high. I would probably start at 10?

momento/src/commands/cloud_linter/dynamodb.rs
```diff
@@ -218,19 +276,11 @@ async fn list_table_names(
     Ok(table_names)
 }
 
-async fn fetch_ddb_resources(
+async fn fetch_ddb_ttl(
```
cprice404, line 279: nit - I was expecting this to return a ttl. maybe rename this fn to `should_fetch_ddb_ttl`?

momento/src/commands/cloud_linter/linter_cli.rs
```diff
@@ -28,17 +30,32 @@ pub async fn run_cloud_linter(region: String) -> Result<(), CliError> {
...
+
+    let describe_ttl_quota = Quota::per_second(
+        core::num::NonZeroU32::new(1).expect("should create non-zero describe_ttl_quota"),
+    );
+    let describe_ttl_limiter = Arc::new(RateLimiter::direct(describe_ttl_quota));
 
-    let mut resources = get_ddb_resources(&config, Arc::clone(&limiter)).await?;
+    let metrics_quota =
+        Quota::per_second(core::num::NonZeroU32::new(20).expect("should create non-zero quota"));
+    let metrics_limiter = Arc::new(RateLimiter::direct(metrics_quota));
```
cprice404, line 45: i like this stuff

cprice404 (commented):
couple of minor questions/nits but overall lgtm

momento/src/commands/cloud_linter/dynamodb.rs
eaddingtonwhite: yeah agree would start low. Originally this was 1000 and i suggested 100. Im down to start at 10 though.

PR #314, opened 2024-04-18
feat: stream resources to files

bruuuuuuuce:
This pr updates the cloud linter command to be able to stream resources to a file, rather than keep everything in memory, before writing to a file

This main difference in this pr is that the entry point just calls `process_resource_<resource_name>`, and then the resource is in charge of getting all metadata/metrics, and then streaming the results to a file. Previously, the `process_resource` function would return the resources for further processing

### DIFF
--- a/momento/Cargo.toml
+++ b/momento/Cargo.toml
@@ -25,6 +25,8 @@ aws-sdk-dynamodb = "1.19.0"
 aws-sdk-elasticache = "1.18.0"
 indicatif = "0.17.8"
 flate2 = "1.0.28"
+json-writer = "0.3.0"
+struson = { version = "0.5.0", features = ["serde"] }
 
 [dev-dependencies]
 assert_cmd = "2.0.2"
@@ -63,7 +65,7 @@ version = "1"
 features = [ "full",]
 
 [dependencies.serde]
-version = "1.0"
+version = "1.0.198"
 features = [ "derive",]
 
 [dependencies.serde_json]
--- a/momento/Cargo.toml
+++ b/momento/Cargo.toml
@@ -25,7 +25,6 @@ aws-sdk-dynamodb = "1.19.0"
 aws-sdk-elasticache = "1.18.0"
 indicatif = "0.17.8"
 flate2 = "1.0.28"
-json-writer = "0.3.0"
 struson = { version = "0.5.0", features = ["serde"] }
 
 [dev-dependencies]
--- a/momento/src/commands/cloud_linter/dynamodb.rs
+++ b/momento/src/commands/cloud_linter/dynamodb.rs
@@ -84,6 +84,8 @@ pub(crate) struct DynamoDbMetadata {
     ttl_enabled: bool,
     #[serde(rename = "isGlobalTable")]
     is_global_table: bool,
+    #[serde(rename = "deleteProtectionEnabled")]
+    delete_protection_enabled: bool,
     #[serde(rename = "lsiCount")]
     lsi_count: i64,
     #[serde(rename = "tableClass")]
@@ -365,6 +367,7 @@ async fn process_table_resources(
         p_throughput_read_units,
         p_throughput_write_units,
         gsi: None,
+        delete_protection_enabled: table.deletion_protection_enabled.unwrap_or_default(),
     };
 
     let mut resources = table
--- a/momento/src/commands/cloud_linter/dynamodb.rs
+++ b/momento/src/commands/cloud_linter/dynamodb.rs
@@ -455,7 +455,9 @@ async fn process_table_resources(
         sender
             .send(Resource::DynamoDb(resource))
             .await
-            .expect("TODO: panic message");
+            .map_err(|err| CliError {
+                msg: format!("Failed to stream dynamodb resource to file: {}", err),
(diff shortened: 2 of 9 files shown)


momento/src/commands/cloud_linter/dynamodb.rs
```diff
@@ -164,91 +167,41 @@ impl ResourceWithMetrics for DynamoDbResource {
...
+        ProgressBar::new(table_names.len() as u64).with_message("Processing Dynamo DB tables");
     describe_ddb_tables_bar
         .set_style(ProgressStyle::with_template("  {msg} {bar} {eta}").expect("invalid template"));
-    let mut ddb_resources: Vec<DynamoDbResource> = Vec::new();
     for table_name in table_names {
-        let instances = fetch_ddb_resources(&ddb_client, &table_name, Arc::clone(&limiter)).await?;
-        for instance in instances {
-            ddb_resources.push(instance);
-        }
+        process_table_resources(
```
bruuuuuuuce, line 190: here we could probably just launch the task in the background, rather than awaiting?

momento/Cargo.toml
```diff
@@ -25,6 +25,8 @@ aws-sdk-dynamodb = "1.19.0"
 aws-sdk-elasticache = "1.18.0"
 indicatif = "0.17.8"
 flate2 = "1.0.28"
+json-writer = "0.3.0"
```
bruuuuuuuce, line 28:
```suggestion
```

momento/src/commands/cloud_linter/elasticache.rs
```diff
@@ -232,5 +241,19 @@ async fn convert_to_resources(
             }
         };
     }
-    Ok(resources)
+
+    for resource in resources {
+        match resource {
+            Resource::DynamoDb(_) => todo!(),
```
cprice404, line 247: que?

momento/src/commands/cloud_linter/linter_cli.rs
```diff
@@ -1,87 +1,95 @@
...
+use super::elasticache::process_elasticache_resources;
+use super::resource::Resource;
 
 pub async fn run_cloud_linter(region: String) -> Result<(), CliError> {
+    let (tx, mut rx) = mpsc::channel::<Resource>(32);
+    let file_path = "linter_results.json";
+    // first we check to make sure we have perms to write files to the current directory
+    check_output_is_writable(file_path).await?;
+
+    // here we right the unzipped json file, containing all the linter results
```
cprice404, line 26:
```suggestion
    // here we write the unzipped json file, containing all the linter results
```

momento/src/commands/cloud_linter/linter_cli.rs
```diff
@@ -1,87 +1,95 @@
...
+    });
+    while let Some(message) = rx.recv().await {
+        let _ = json_writer.serialize_value(&message);
+    }
+    json_writer.end_array()?;
+    json_writer.end_object()?;
+    json_writer.finish_document()?;
+
+    // now we compress the json into a .gz file for the customer to upload
+    let opened_file_tokio = File::open(file_path).await?;
```
cprice404, line 44: this is a silly nit but i don't understand why these file objects all have `tokio` in their var name :)

nand4011: There are separate tokio versions of a lot of the file handling stuff. Better for async I guess.

momento/src/commands/cloud_linter/elasticache.rs
bruuuuuuuce: O yea this is a case that should never happen since we are only processing elasticache resources. I can clear this up and just `continue` if this happens

momento/src/commands/cloud_linter/linter_cli.rs
bruuuuuuuce: Tokio has its own version of all the file system stuff. I could not get the json streaming library to work with the tokio versions of files, since they don't implement `std::io::write` trait they instead implement the tokio version of this trait. There may be an easy work around for this, but I time boxed it and could not get it to work. So here we read the files using tokio, and then we convert them to `std` files so we can stream the json data to them

momento/src/commands/cloud_linter/dynamodb.rs
```diff
@@ -460,5 +442,18 @@ async fn fetch_ddb_resources(
...
+        resource
+            .append_metrics(metrics_client, Arc::clone(&metrics_limiter))
+            .await?;
+        let ttl_enabled =
+            is_ddb_ttl_enabled(ddb_client, &resource, Arc::clone(&describe_ttl_limiter)).await?;
+        resource.metadata.ttl_enabled = ttl_enabled;
+        sender
+            .send(Resource::DynamoDb(resource))
+            .await
+            .expect("TODO: panic message");
```
cprice404, line 455: you sent me a DM about this LOC the other day. I think I understand the reason it won't compile with a `?`. lmk if you want to discuss/pair on it.

nand4011 (approved):
LGTM unless we want to get rid of the todos.

nand4011 (approved):

PR #317, opened 2024-04-19
fix: more occurrences of ENV:GITHUB_OUTPUT, and better error if rpm/deb files missing

cprice404:


.github/workflows/execute-release.yml
```diff
@@ -54,7 +54,7 @@ jobs:
 
       - name: Output release
         id: release
-        run: echo "release=${{ steps.semrel.outputs.version }}" >> $GITHUB_OUTPUT
+        run: echo "release=${{ steps.semrel.outputs.version }}" >> $ENV:GITHUB_OUTPUT
```
bruuuuuuuce, line 57: how does it work here https://github.com/momentohq/client-sdk-javascript/blob/main/.github/workflows/on-push-to-release-branch.yml#L30?

cprice404: i don't know. all i know is that in the failures that I saw from this repo, the places where we were doing `ENV:GITHUB_OUTPUT` seemed to work, and these ones seem to not be working, so I'm just throwing a dart

bruuuuuuuce (approved):
:shipit:

PR #318, opened 2024-04-19
fix: still working on GITHUB_OUTPUT

cprice404:
Now I'm thinking that all the ones that happen on Windows need
to be `$ENV:GITHUB_OUTPUT` and the ones that happen on mac/linux
need to be just `$GITHUB_OUTPUT`.


bruuuuuuuce (approved):
:shipit:

PR #320, opened 2024-04-24
feat: serverless elasticache

bruuuuuuuce:


momento/src/commands/cloud_linter/serverless_elasticache.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    write_resources(clusters, config, region, sender, metrics_limiter).await?;
+    Ok(())
+}
+
+async fn describe_clusters(
+    elasticache_client: &aws_sdk_elasticache::Client,
+    limiter: Arc<DefaultDirectRateLimiter>,
+) -> Result<Vec<ServerlessCache>, CliError> {
+    let bar =
```
nand4011, line 141: I think we should look into using a [MultiProgress bar](https://docs.rs/indicatif/latest/indicatif/struct.MultiProgress.html) since we're processing in parallel now, but that can be an improvement if we decide to make any more changes.

nand4011 (approved):

PR #325, opened 2024-05-30
feat: speed up resource collection, add resource specific cli flags

bruuuuuuuce:
closes https://github.com/momentohq/dev-eco-issue-tracker/issues/856

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,7 @@ client-cli.iml
 bin/
 momento.exe
 obj/
+
+linter_results.json
+linter_results.json.gz
+
--- a/momento-cli-opts/src/lib.rs
+++ b/momento-cli-opts/src/lib.rs
@@ -224,7 +224,32 @@ to help find opportunities for optimizations with Momento.
     CloudLinter {
         #[arg(long, short, help = "The AWS region to examine")]
         region: String,
+        #[arg(
+            long = "enable-ddb-ttl-check",
+            help = "Opt in to check whether ddb tables have ttl enabled. If there are lots of tables, could slow down data collection"
+        )]
+        enable_ddb_ttl_check: bool,
+        #[arg(
+            value_enum,
+            long = "resource",
+            help = "Pass in a specific resource type to only collect data on that resource. Example: --resource dynamo"
+        )]
+        resource: Option<CloudLinterResources>,
+        #[arg(
+            long = "metric-collection-rate",
+            help = "tps at which to invoke the aws `get-metric-data` api",
+            default_value = "20"
+        )]
+        metric_collection_rate: u32,
     },
+}
+
+#[derive(clap::ValueEnum, PartialEq, Eq, Debug, Clone, Copy)]
+pub enum CloudLinterResources {
+    ApiGateway,
+    S3,
+    Dynamo,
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -7,3 +7,7 @@ client-cli.iml
 bin/
 momento.exe
 obj/
+
+linter_results.json
+linter_results.json.gz
```
pgautier404, line 12: Thank you.

momento-cli-opts/src/lib.rs
```diff
@@ -224,9 +224,34 @@ to help find opportunities for optimizations with Momento.
...
+            long = "enable-ddb-ttl-check",
+            help = "Opt in to check whether ddb tables have ttl enabled. If there are lots of tables, could slow down data collection"
+        )]
+        enable_ddb_ttl_check: bool,
+        #[arg(
+            value_enum,
+            long = "resource",
+            help = "Pass in a specific resource type to only collect data on that resource. Example: --resource dynamo"
+        )]
+        resource: Option<CloudLinterResources>,
```
pgautier404, line 237: It sounded to me like this should accept multiple values. @eaddingtonwhite is that right?

momento/src/commands/cloud_linter/dynamodb.rs
```diff
@@ -451,8 +484,13 @@ async fn process_table_resources(
...
             .append_metrics(metrics_client, Arc::clone(&metrics_limiter))
             .await?;
-        let ttl_enabled =
-            is_ddb_ttl_enabled(ddb_client, &resource, Arc::clone(&describe_ttl_limiter)).await?;
+        let ttl_enabled = match enable_ddb_ttl_check {
+            true => {
+                is_ddb_ttl_enabled(ddb_client, &resource, Arc::clone(&describe_ttl_limiter)).await?
+            }
+            false => false,
+        };
```
pgautier404, line 492: This doesn't allow us to discriminate between the ttl check itself being disabled and ttl being disabled on all tables. I'm not sure that's a distinction we need to be able to make, but if it is we'll need to add the command line options to the output or otherwise make that flag's setting available.

bruuuuuuuce: This is a good point, I think what we currently has also has a similar problem where, we use a heuristic to determine if we should call that describe ttl api, but if we don't call the api, we just mark the ttl as false, and don't mark anywhere that we never tried to lookup the ttl. I think we should probably add a field to the json that is something along the lines of `attempted_to_get_ttl`

momento-cli-opts/src/lib.rs
eaddingtonwhite: Yeah that was my original ask but its ok if just limits to one for now. @bruuuuuuuce can you make follow up ticket to allow passing a list here. We can fix in follow up pr.

momento/src/commands/cloud_linter/dynamodb.rs
bruuuuuuuce: I think the backend is expecting it to be a boolean so we would need to make some changes there before sending a non-boolean value

pgautier404 (approved):

PR #326, opened 2024-06-03
fix: more parrellelization around dynamodb metric collection

bruuuuuuuce:


momento/src/commands/cloud_linter/dynamodb.rs
```diff
@@ -179,58 +213,68 @@ pub(crate) async fn process_ddb_resources(
...
     enable_ddb_ttl_check: bool,
+    enable_gsi: bool,
 ) -> Result<(), CliError> {
     let ddb_client = aws_sdk_dynamodb::Client::new(config);
     let metrics_client = aws_sdk_cloudwatch::Client::new(config);
 
     let list_ddb_tables_bar = ProgressBar::new_spinner().with_message("Listing Dynamo DB tables");
     list_ddb_tables_bar.enable_steady_tick(Duration::from_millis(100));
-    let table_names = list_table_names(&ddb_client, Arc::clone(&control_plane_limiter)).await?;
+    let table_names = list_table_names(&ddb_client, Arc::clone(&metrics_limiter)).await?;
```
cprice404, line 223: not a blocker for this PR but curious why this limiter is changing and if we even need to keep the control plane one

momento/src/commands/cloud_linter/linter_cli.rs
```diff
@@ -73,11 +77,17 @@ async fn process_data(
...
     sender: Sender<Resource>,
     enable_ddb_ttl_check: bool,
+    enable_gsi: bool,
     only_collect_for_resource: Option<CloudLinterResources>,
     metric_collection_rate: u32,
 ) -> Result<(), CliError> {
+    let retry_config = RetryConfig::adaptive()
+        .with_initial_backoff(Duration::from_secs(1))
+        .with_max_attempts(20)
+        .with_max_backoff(Duration::from_secs(15));
```
cprice404, line 87: Also not a blocker for this PR but how much visibility do we get into the retries that are happening with this? I almost wonder if we should just roll our own simple retry logic so that we can more easily debug what's happening.

momento/src/commands/cloud_linter/metrics.rs
```diff
@@ -45,6 +45,29 @@ pub(crate) trait AppendMetrics {
     ) -> Result<(), CliError>;
 }
 
+// impl<T> AppendMetrics for T
```
cprice404, line 48: not a blocker but what's up with this

momento/src/commands/cloud_linter/metrics.rs
```diff
@@ -153,7 +176,16 @@ async fn query_metrics_for_target(
...
+            let result = match result {
+                Ok(res) => res,
+                Err(e) => {
+                    println!("get_metric_data_error: {:?}", e);
+                    return Err(CliError {
+                        msg: "error from aws api while querying metrics".to_string(),
+                    });
+                }
+            };
+            // let result = result?;
```
cprice404, line 188: errant comment?

cprice404 (approved):
code looks good to me, left a handful of questions / nits but none are blockers

momento/src/commands/cloud_linter/dynamodb.rs
bruuuuuuuce: o you are right this should be control plane limiter...

bruuuuuuuce, line 223:
```suggestion
    let table_names = list_table_names(&ddb_client, Arc::clone(&control_plane_limiter)).await?;
```

momento/src/commands/cloud_linter/linter_cli.rs
bruuuuuuuce: yea thats not a bad idea, I spent a bunch of time trying to tune the retries

momento/src/commands/cloud_linter/metrics.rs
bruuuuuuuce, line 188:
```suggestion
```

cprice404 (approved):

PR #327, opened 2024-06-03
feat: speed up resource collection for elasticache

bruuuuuuuce:
image

### DIFF
--- a/momento/src/commands/cloud_linter/elasticache.rs
+++ b/momento/src/commands/cloud_linter/elasticache.rs
@@ -4,8 +4,9 @@ use std::time::Duration;
 
 use aws_config::SdkConfig;
 use aws_sdk_elasticache::types::CacheCluster;
+use futures::stream::FuturesUnordered;
 use governor::DefaultDirectRateLimiter;
-use indicatif::ProgressBar;
+use indicatif::{ProgressBar, ProgressStyle};
 use phf::{phf_map, Map};
 use serde::Serialize;
 use tokio::sync::mpsc::Sender;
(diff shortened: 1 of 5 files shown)


momento/src/commands/cloud_linter/elasticache.rs
```diff
@@ -116,30 +117,99 @@ pub(crate) async fn process_elasticache_resources(
...
+        &metrics_client,
+        control_plane_limiter,
+        metrics_limiter,
+        region,
+        sender,
+    )
+    .await?;
 
-    write_resources(clusters, config, region, sender, metrics_limiter).await?;
+    // write_resources(clusters, config, region, sender, metrics_limiter).await?;
```
cprice404, line 131: ?

bruuuuuuuce: I can remove that, writing to the file is now handled in the `process_resources` function

bruuuuuuuce, line 131:
```suggestion
```

momento/src/commands/cloud_linter/elasticache.rs
```diff
@@ -116,30 +117,99 @@ pub(crate) async fn process_elasticache_resources(
...
+            // For more spinners check out the cli-spinners project:
+            // https://github.com/sindresorhus/cli-spinners/blob/master/spinners.json
+            .tick_strings(&[
+                "▹▹▹▹▹",
+                "▸▹▹▹▹",
+                "▹▸▹▹▹",
+                "▹▹▸▹▹",
+                "▹▹▹▸▹",
+                "▹▹▹▹▸",
+                "▪▪▪▪▪",
```
cprice404, line 156: why this instead of a count like you did in the other ones?

cprice404 (commented):
looks okay to me, just the question about why we would style this progress bar differently than the others

momento/src/commands/cloud_linter/elasticache.rs
bruuuuuuuce: With s3/dynamo, we can list all of the tables/bucket names, and then for each table/bucket we can call the `describe` api. With elasticache, we get back an iterator over all the clusters. We don't know how many clusters their are. To use the progress bar like we did with s3/dynamo, we need to know the total number of resources we are querying over, and then incr the progress bar at the per resource level. Since we don't know how many resources we are going to be querying over, we need to use the `spinner` feature, which shows a loader and a counter of resources that we have processed so far. There may be a way to have a dynamic progress bar, but I could not getting that working

cprice404: this sounds like an improvement so i'm not necessarily saying we should block this PR for it, but couldn't we just run over the iterator once and accumulate all the clusters into a list, and then loop over that to make the metrics API calls?

bruuuuuuuce: yea thats fair we could

cprice404 (approved):

PR #328, opened 2024-06-05
chore: improve progress bars in cloud-linter

nand4011:
Separate the metrics calls from the describe tables calls for elasticache and serverless elasticache. This lets us use a spinner for the describe calls, and then a standard progress bar for the metrics calls.

Remove an extra call to increment the s3 progress bar.

Add a progress bar to display that we are writing out the results file.

Misc linter fixes.

### DIFF
--- a/momento/src/commands/cloud_linter/s3.rs
+++ b/momento/src/commands/cloud_linter/s3.rs
@@ -243,7 +243,7 @@ async fn process_buckets(
     control_plane_limiter: Arc<DefaultDirectRateLimiter>,
 ) -> Result<(), CliError> {
     let process_buckets_bar =
-        ProgressBar::new((buckets.len()) as u64).with_message("Processing S3 Buckets");
+        ProgressBar::new(buckets.len() as u64).with_message("Processing S3 Buckets");
     process_buckets_bar.set_style(
         ProgressStyle::with_template(" {pos:>7}/{len:7} {msg}").expect("invalid template"),
     );
@@ -258,7 +258,6 @@ async fn process_buckets(
         let region_clone = region.to_string().clone();
         let progress_bar_clone = process_buckets_bar.clone();
         let spawn = tokio::spawn(async move {
-            progress_bar_clone.inc(1);
             let res = process_bucket(
                 s3_client_clone,
                 bucket,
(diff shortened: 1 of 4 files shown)


momento/src/commands/cloud_linter/s3.rs
```diff
@@ -258,7 +258,6 @@ async fn process_buckets(
         let region_clone = region.to_string().clone();
         let progress_bar_clone = process_buckets_bar.clone();
         let spawn = tokio::spawn(async move {
-            progress_bar_clone.inc(1);
```
cprice404, line 261: too funny

cprice404 (approved):
dope

Let's let @bruuuuuuuce look too tho

momento/src/commands/cloud_linter/serverless_elasticache.rs
```diff
@@ -152,24 +151,43 @@ async fn process_resources(
...
+    let resources = describe_caches(elasticache_client, control_plane_limiter, region).await?;
+    describe_bar.finish();
+
+    let process_bar = ProgressBar::new(resources.len() as u64)
+        .with_message("Processing Serverless ElastiCache resources");
+    process_bar.set_style(
+        ProgressStyle::with_template(" {pos:>7}/{len:7} {msg}").expect("invalid template"),
     );
+
+    for mut resource in resources {
```
bruuuuuuuce, line 166: both here and [here](https://github.com/momentohq/momento-cli/pull/328/files#diff-d8cd92ad553404856ec28e5d93dab25c5476da8486f06bceb4bca6821221e618R152), we are going back to linearly fetching metrics for each resource. I think if here instead of looping through each resource 1 by 1, you spawned like 10 resources at a time, and had them all query metrics in parallel, it would get the metrics at a much faster rate. Similar to how we had it [here](https://github.com/momentohq/momento-cli/pull/328/files#diff-d8cd92ad553404856ec28e5d93dab25c5476da8486f06bceb4bca6821221e618L178-L209)

momento/src/commands/cloud_linter/serverless_elasticache.rs
```diff
@@ -163,19 +164,49 @@ async fn process_resources(
         ProgressStyle::with_template(" {pos:>7}/{len:7} {msg}").expect("invalid template"),
     );
 
-    for mut resource in resources {
-        resource
-            .append_metrics(metrics_client, Arc::clone(&metrics_limiter))
-            .await?;
+    while !resources.is_empty() {
+        let chunk: Vec<ServerlessElastiCacheResource> = resources
```
bruuuuuuuce, line 168: i like this way better than how I was doing this :fistbump:. Not for this pr, but would be nice to have a higher level function around this, to just DRY up the code around this

bruuuuuuuce (approved):
:shipit:

momento/src/commands/cloud_linter/serverless_elasticache.rs
nand4011: Agreed.

PR #329, opened 2024-08-01
feat: Make the cloud linter start and end dates configurable

nand4011:
Add optional arguments to the cloud-linter command that take a start and end date in YYYY-MM-DD form. If the start-date isn't provided, it defaults to end-date - 30. If end-date isn't provided, it defaults to today.

Make the process resources function argument orders consistent.

Delete some commented out code.

### DIFF
--- a/momento-cli-opts/src/lib.rs
+++ b/momento-cli-opts/src/lib.rs
@@ -258,12 +258,12 @@ to help find opportunities for optimizations with Momento.
         metric_collection_rate: u32,
         #[arg(
             long = "start-date",
-            help = "The UTC start date of the metric collection period. Will use (end-date - 30 days) if not provided. (YYYY-MM-DD)"
+            help = "The inclusive UTC start date of the metric collection period. Will use (end-date - 30 days) if not provided. (YYYY-MM-DD)"
         )]
         metric_start_date: Option<String>,
         #[arg(
             long = "end-date",
-            help = "The UTC end date of the metric collection period. Will use the current date if not provided. (YYYY-MM-DD)"
+            help = "The inclusive UTC end date of the metric collection period. Will use the current date if not provided. (YYYY-MM-DD)"
         )]
         metric_end_date: Option<String>,
     },
--- a/momento-cli-opts/src/lib.rs
+++ b/momento-cli-opts/src/lib.rs
@@ -256,6 +256,16 @@ to help find opportunities for optimizations with Momento.
             default_value = "10"
         )]
         metric_collection_rate: u32,
+        #[arg(
+            long = "start-date",
+            help = "The UTC start date of the metric collection period. Will use (end-date - 30 days) if not provided. (YYYY-MM-DD)"
+        )]
+        metric_start_date: Option<String>,
+        #[arg(
+            long = "end-date",
+            help = "The UTC end date of the metric collection period. Will use the current date if not provided. (YYYY-MM-DD)"
+        )]
+        metric_end_date: Option<String>,
     },
 }
 
--- a/momento/src/commands/cloud_linter/linter_cli.rs
+++ b/momento/src/commands/cloud_linter/linter_cli.rs
@@ -25,6 +25,7 @@ use crate::error::CliError;
 use super::elasticache::process_elasticache_resources;
 use super::resource::Resource;
 
+#[allow(clippy::too_many_arguments)]
 pub async fn run_cloud_linter(
     region: String,
     enable_ddb_ttl_check: bool,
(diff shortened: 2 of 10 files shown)


momento/src/commands/cloud_linter/linter_cli.rs
```diff
@@ -210,12 +237,65 @@ async fn process_data(
...
+/// A Result containing a tuple of the start and end timestamps in millis, or a CliError
+/// if date parsing fails.
+fn get_metric_time_range(
+    start_date: Option<String>,
+    end_date: Option<String>,
+) -> Result<(i64, i64), CliError> {
+    let now = Utc::now();
+    let thirty_days = chrono::Duration::days(30);
+
+    let parse_date = |date_str: &str| -> Result<NaiveDateTime, CliError> {
```
cprice404, line 268: just curious - why is this a lambda rather than a regular `fn`?

momento/src/commands/cloud_linter/linter_cli.rs
```diff
@@ -210,12 +237,65 @@ async fn process_data(
...
+        let date = NaiveDate::parse_from_str(date_str, "%Y-%m-%d")?;
+        date.and_hms_opt(0, 0, 0).ok_or_else(|| CliError {
+            msg: "invalid time".to_string(),
+        })
+    };
+
+    match (start_date.as_deref(), end_date.as_deref()) {
```
cprice404, line 275: super nit but I feel like you could just do something like:

```rust
let processed_end_date = end_date.map_or(now.timestamp_millis(), parse_date);
let processed_start_date = start_date.map_or((end - thirty_days).timestamp_millis(), parse_date);
```

And then you wouldn't need this big matrix match statement.

cprice404 (commented):
couple of minor questions / nits that probably aren't blockers.

We may want @eaddingtonwhite to pull down and run it before merging, if he has time.

momento/src/commands/cloud_linter/linter_cli.rs
nand4011: It wasn't needed outside of the function. I don't have a problem moving it out.

momento/src/commands/cloud_linter/linter_cli.rs
nand4011: That makes sense. I was dealing with all 4 cases while I decided what the behavior should be if the user didn't provided start_date or end_date. Originally I was going to do end_date = (start_date + 30) instead of end_date = now. Your solution is much nicer.

nand4011:
This won't pass the linter until https://github.com/momentohq/momento-cli/pull/330 is merged

momento-cli-opts/src/lib.rs
```diff
@@ -256,6 +256,16 @@ to help find opportunities for optimizations with Momento.
...
+        #[arg(
+            long = "start-date",
+            help = "The UTC start date of the metric collection period. Will use (end-date - 30 days) if not provided. (YYYY-MM-DD)"
+        )]
+        metric_start_date: Option<String>,
+        #[arg(
+            long = "end-date",
+            help = "The UTC end date of the metric collection period. Will use the current date if not provided. (YYYY-MM-DD)"
+        )]
+        metric_end_date: Option<String>,
```
malandis, line 268: Are the dates inclusive/exclusive? It's worth making this clear in the argument documentation.

momento/src/commands/cloud_linter/linter_cli.rs
```diff
@@ -221,12 +249,57 @@ async fn process_data(
...
+        .transpose()?
+        .map(|date| date.timestamp_millis())
+        .unwrap_or_else(|| processed_end_date - thirty_days);
+
+    Ok((processed_start_date, processed_end_date))
+}
+
+fn parse_date_string(date: &str) -> Result<NaiveDateTime, CliError> {
+    let naive_date = NaiveDate::parse_from_str(date, "%Y-%m-%d").map_err(|_| CliError {
+        msg: "Date must be in ISO 8601 (YYYY-MM-DD) format".to_string(),
```
malandis, line 296:
```suggestion
        msg: "Date must be in YYYY-MM-DD format".to_string(),
```

The `YYYY-MM-DD` is more commonplace than the ISO specification number hence more user friendly.

momento-cli-opts/src/lib.rs
nand4011: It already specifies UTC. I can try to make it the timezone of the region if that is desired. I'll add that the dates are inclusive.

malandis: Thanks! I missed the UTC bit.

malandis (approved):

PR #339, opened 2025-07-03
feat: upgrade momento dependency, remove Login, SigningKeys, and GenerateToken commands

anitarua:
The Momento Rust SDK version in the CLI has not been updated in a long time. We've since revamped the Rust SDK and the APIs used for the Login, SigningKeys, and GenerateToken commands are no longer available. These commands were under the Preview category and the CLI is not yet v1.0 so opting to just remove them.


momento/src/commands/topic/mod.rs
```diff
@@ -1,19 +1,14 @@
...
-                momento::preview::topics::ValueKind::Binary(binary) => {
-                    console_data!("{{\"kind\": \"binary\", \"length\": {}}}", binary.len())
-                }
-            },
-            momento::preview::topics::SubscriptionItem::Discontinuity(discontinuity) => {
-                console_data!("{discontinuity:?}")
+        match item.kind {
+            momento::topics::ValueKind::Text(text) => console_data!("{text}"),
+            momento::topics::ValueKind::Binary(binary) => {
+                console_data!("{{\"kind\": \"binary\", \"length\": {}}}", binary.len())
```
malandis, line 11: Do we not have display or other such traits defined for the messages?

malandis (dismissed):
LGTM just:

- Have you manually tested the affected commands due the upgrade? We don't have test coverage for printing etc.

In the future let's have separate PRs for the feature removal vs client upgrade. Not a big deal in this case helps make review that much easier 😄

momento/src/commands/topic/mod.rs
anitarua: We have Debug implemented for these, will update

malandis (approved):

PR #340, opened 2025-07-08
feat: add put/list wasms and put/list functions

anitarua:
Adds these methods to the CLI under the `preview function` subcommands:

- PutFunction
- ListFunctions
- ListFunctionVersions
- PutWasm
- ListWasms

Open to different naming and UX suggestions! Current state:

- PutFunction takes different optional arguments to determine the WasmSource (uploaded inline or reference an uploaded wasm), but could split that into two different commands if preferable.
- ~PutFunction accepts environment variables as a list, e.g. "key,value,key,value,..."~ REVISED (thanks, Kenny!)
- ~ListWasms and ListFunctionVersions are using debug printouts for now because the returned objects have no field accessors.~
    - ~Here's the [PR in rust SDK](https://github.com/momentohq/client-sdk-rust/pull/469) for the additional accessors, but we should still decide how much information we want to display on each of the wasms/functions/versions.~ DONE

Test commands
```bash
# PutWasm
cargo run -- preview function put-wasm --name <name> --wasm-file <path>

# ListWasms
cargo run -- preview function list-wasms

# PutFunction using wasm file
cargo run -- preview function put-function --name <name> --cache-name <cache> --wasm-file <path>

# PutFunction using uploaded wasm
cargo run -- preview function put-function --name <name> --cache-name <cache> --id-uploaded-wasm <id> --version-uploaded-wasm <version number>

# ListFunctions
cargo run -- preview function list-functions --cache-name <cache>

# ListFunctionVersions
cargo run -- preview function list-function-versions --id <id>
```

### DIFF
--- a/momento-cli-opts/src/lib.rs
+++ b/momento-cli-opts/src/lib.rs
@@ -1,3 +1,5 @@
+use std::error::Error;
+
 use clap::CommandFactory;
 use clap::Parser;
 
@@ -138,10 +140,11 @@ pub enum FunctionCommand {
         description: Option<String>,
         #[arg(
             long = "env-var",
-            short,
-            help = "Environment variables to provide to the Function. Example: --env-var KEY1,VALUE1,KEY2,VALUE2"
+            short = 'E',
+            value_parser = parse_env::<String, String>,
+            help = "Environment variables to provide to the Function. Example: -E KEY1=value_1 -E KEY2=value_2"
         )]
-        environment_variables: Option<Vec<String>>,
+        environment_variables: Vec<(String, String)>,
     },
     #[command(about = "Create or update a wasm source that can be used in a Momento Function")]
     PutWasm {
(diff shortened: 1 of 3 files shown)


momento-cli-opts/src/lib.rs
```diff
@@ -131,6 +108,68 @@ pub enum AccountCommand {
...
+        )]
+        version_uploaded_wasm: Option<u32>,
+        #[arg(long = "description", short, help = "Description")]
+        description: Option<String>,
+        #[arg(
+            long = "env-var",
+            short,
+            help = "Environment variables to provide to the Function. Example: --env-var KEY1,VALUE1,KEY2,VALUE2"
+        )]
+        environment_variables: Option<Vec<String>>,
```
kvcache, line 144: you could instead use a parser

```rust
fn parse_env<K, V>(s: &str) -> Result<(K, V), Box<dyn Error + Send + Sync + 'static>>
where
    K: std::str::FromStr,
    K::Err: Error + Send + Sync + 'static,
    V: std::str::FromStr,
    V::Err: Error + Send + Sync + 'static,
{
    let equals = s
        .find('=')
        .ok_or_else(|| format!("invalid KEY=value: no `=` found in `{s}`"))?;
    let key = s[..equals].parse()?;
    let value = s[pos + 1..].parse()?;
    Ok((key, value))
}
```

and then it's
```rust
        #[arg(
            long = "env-var",
            short = "E",
            value_parser = parse_env::<String, String>,
            help = "Environment variables to provide to the Function. Example: -E KEY1=value_1 -E KEY2=value_2"
        )]
        environment_variables: Vec<(String, String)>,
```

This is error-resistant. Instead of depending on people to align their list of pairs, it relates the pairs at the type level by the arguments

anitarua: oh nice! will update to use that instead, thanks!

kvcache (approved):
looks great, thank you!

